PR #7, opened 2025-06-17
avoid pydantic deprecations

AstraBert:
I switched from the deprecated `model_fields` in Pydantic BaseModels to `__class__.model_fields`

### DIFF
--- a/src/workflows/events.py
+++ b/src/workflows/events.py
@@ -69,7 +69,6 @@ class Event(BaseModel):
         private_attrs = {}
         data = {}
         for k, v in params.items():
-            # using __class__.model_fields because using model_fields directly is deprecated
             if k in self.__class__.model_fields:
                 fields[k] = v
             elif k in self.__private_attributes__:


src/workflows/events.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, **params: Any):
         private_attrs = {}
         data = {}
         for k, v in params.items():
-            if k in self.model_fields:
+            if k in self.__pydantic_fields__:
```
logan-markewich: I think `self.__class__.model_fields` also works, not sure what we prefer though

AstraBert: MMhh I think they are suggesting your way of doing it in their deprecation note

logan-markewich (approved):

src/workflows/events.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, **params: Any):
         private_attrs = {}
         data = {}
         for k, v in params.items():
-            if k in self.model_fields:
+            # using __class__.model_fields because using model_fields directly is deprecated
```
masci: Nit: ironically this comment will keep the deprecated method mentioned forever 😅 it should be enough to look at the commit message for future reference

masci:
```suggestion
```

masci (approved):

PR #8, opened 2025-06-17
chore: add readme + contributing

logan-markewich:
Adding an initial README -- feel free to suggest anything else! I kind of guessed about what we want in here

Also spun up a quick contributing.md to make the repo a bit more dev-friendly :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ LlamaIndex Workflows are a framework for orchestrating and chaining together com
 Workflows shine when you need to orchestrate complex, multi-step processes that involve AI models, APIs, and decision-making. Here are some examples of what you can build:
 
 - **AI Agents** - Create intelligent systems that can reason, make decisions, and take actions across multiple steps
-- **Document Processing Pipelines** - Build systems that ingest, analyze, summarize, and route documents through various processing stages  
+- **Document Processing Pipelines** - Build systems that ingest, analyze, summarize, and route documents through various processing stages
 - **Multi-Model AI Applications** - Coordinate between different AI models (LLMs, vision models, etc.) to solve complex tasks
 - **Research Assistants** - Develop workflows that can search, analyze, synthesize information, and provide comprehensive answers
 - **Content Generation Systems** - Create pipelines that generate, review, edit, and publish content with human-in-the-loop approval


README.md
```diff
@@ -0,0 +1,75 @@
...
+
+[![PyPI - Downloads](https://img.shields.io/pypi/dm/llama-index)](https://pypi.org/project/llama-index-workflows/)
+[![GitHub contributors](https://img.shields.io/github/contributors/run-llama/llama-index-workflows)](https://github.com/run-llama/llama-index-workflows/graphs/contributors)
+[![Discord](https://img.shields.io/discord/1059199217496772688)](https://discord.gg/dGcwcsnxhU)
+[![Twitter](https://img.shields.io/twitter/follow/llama_index)](https://x.com/llama_index)
+[![Reddit](https://img.shields.io/reddit/subreddit-subscribers/LlamaIndex?style=plastic&logo=reddit&label=r%2FLlamaIndex&labelColor=white)](https://www.reddit.com/r/LlamaIndex/)
+
+LlamaIndex Workflows are a framework for orchestrating and chaining together complex systems of steps and events.
+
+Using Workflows, you get access to an async-first event-driven system for building and running your own workflows.
```
masci: I would expand this sentence into a paragraph like "What can you build with workflow" or somthing along those lines. Claude suggests something like this:

With LlamaIndex Workflows, you can build sophisticated AI agent systems that coordinate multiple tools, APIs, and reasoning steps. Create agents that can plan and execute complex tasks, handle tool calling sequences, manage conversation context across multiple interactions, and implement advanced patterns like ReAct, planning agents, and multi-agent collaboration. The event-driven architecture makes it easy to build agents that can dynamically route between different capabilities, handle error recovery and retries, and maintain state across long-running agent interactions.

CONTRIBUTING.md
```diff
@@ -0,0 +1,68 @@
...
+- **[src/workflows/utils.py](src/workflows/utils.py)**: Utility functions for step introspection, signature validation, and other helpers
+
+### Setup
+
+This section assumes you have `uv` installed.
+
+The install is simply:
+
+```bash
+uv sync
```
masci, line 55: do we want to suggest contributors to work on an active venv? That way they would get pre-commit hooks for free. We might mention `uv run` as an alternative for when the venv is not active

logan-markewich: Yea I was running into that as well, wasn't sure what the standard was for `uv` and pre-commit hooks. I can rewrite this slightly to account for that

README.md
logan-markewich: Makes sense to me! Let me give it a try

CONTRIBUTING.md
logan-markewich: Added! You actually need `uv run pre-commit install` which I completely forgot about lol

README.md
logan-markewich: Took a stab at this, lemme know what you think

CONTRIBUTING.md
masci: Ah good catch, I always forget about the install step 🤦

PR #15, opened 2025-06-21
Add a typed-context API

logan-markewich:
This PR adds an opt-in typed-context state. This effectively replaces the old get/set API 

By default, the state is effectively an untyped dict (taking advantage of the base `Event` class features). This means users will see no breaking change in core usage, and the existing serdes flow continues to work.

However, users can also opt-in to a typed state experience. 

```python
from pydantic import BaseModel, Field

from workflows import Context, Workflow
from workflows.decorators import step
from workflows.events import StartEvent, StopEvent

# State must have defaults for every field
class MyState(BaseModel):
    name: str = Field(default="Jane")
    age: int = Field(default=25)


class MyWorkflow(Workflow):
    @step
    async def step(self, ctx: Context[MyState], ev: StartEvent) -> StopEvent:
        # Modify state attributes
        await ctx.state.set("name", "John")
        await ctx.state.set("age", 30)

        # Get and update entire state
        state = await ctx.state.get_all()
        state.age += 1
        await ctx.state.set_all(state)

        return StopEvent()


async def main():
    wf = MyWorkflow()

    handler = wf.run()

    # Run the workflow
    _ = await handler

    # Check final state
    state = await handler.ctx.state.get_all()
    assert state.model_dump() == MyState(name="John", age=31).model_dump()
```

Design:
- Async first
- Non-breaking
- Designed to (hopefully) be DB ready at some point in the future if we want remote context

Limitations:
- Only one "state" type is supported per workflow (although we could easily support more than one)
- We might want a run_id or context_id in the future to support remote-storage easily.
- `ctx.clear` might need to be deprecated because its not async 

Deprecations/Breaking Changes
- Removed `ctx.data`, `ctx._serialize_globals`, and `ctx._globals`
- Deprecated `ctx.get()` and `ctx.set()` in favor of `ctx.state.get()` and `ctx.state.set()`

### DIFF
--- a/src/workflows/context/context.py
+++ b/src/workflows/context/context.py
@@ -89,7 +89,7 @@ class Context(Generic[MODEL_T]):
     async def _init_state_store(self, state_class: MODEL_T) -> None:
         # If a state manager already exists, ensure the requested state type is compatible
         if self._state_store is not None:
-            existing_state = await self._state_store.get_all()
+            existing_state = await self._state_store.get_state()
             if type(state_class) is not type(existing_state):
                 # Existing state type differs from the requested one – this is not allowed
                 raise ValueError(
--- a/src/workflows/context/state_store.py
+++ b/src/workflows/context/state_store.py
@@ -90,11 +90,11 @@ class InMemoryStateStore(Generic[MODEL_T]):
         self._state = initial_state
         self._lock = asyncio.Lock()
 
-    async def get_all(self) -> MODEL_T:
+    async def get_state(self) -> MODEL_T:
         """Get a copy of the current state."""
         return self._state.model_copy()
 
-    async def set_all(self, state: MODEL_T) -> None:
+    async def set_state(self, state: MODEL_T) -> None:
         """Set the current state."""
         if not isinstance(state, type(self._state)):
             raise ValueError(f"State must be of type {type(self._state)}")
--- a/tests/test_state_manager.py
+++ b/tests/test_state_manager.py
@@ -64,7 +64,7 @@ async def test_state_manager_defaults(
     default_state_manager: InMemoryStateStore[DictState],
 ) -> None:
     assert (
-        await default_state_manager.get_all()
+        await default_state_manager.get_state()
     ).model_dump_json() == DictState().model_dump_json()
 
     await default_state_manager.set("name", "John")
@@ -79,7 +79,7 @@ async def test_state_manager_defaults(
     await default_state_manager.set("nested.a", "c")
     assert await default_state_manager.get("nested.a") == "c"
 
-    full_state = await default_state_manager.get_all()
+    full_state = await default_state_manager.get_state()
     assert full_state.name == "John"
     assert full_state.age == 30
     assert full_state.nested["a"] == "c"
@@ -90,7 +90,7 @@ async def test_default_state_manager_serialization(
     default_state_manager: InMemoryStateStore[DictState],
 ) -> None:
     assert (
-        await default_state_manager.get_all()
+        await default_state_manager.get_state()
     ).model_dump() == DictState().model_dump()
 
     await default_state_manager.set("name", "John")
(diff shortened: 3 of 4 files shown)


src/workflows/context/context.py
```diff
@@ -79,11 +81,35 @@ def __init__(
...
+            return
+
+        # First-time initialisation
+        self._state_manager = InMemoryStateManager(state_class)
+
+    @property
+    def state(self) -> InMemoryStateManager[MODEL_T]:
+        # Default to DictState if no state manager is initialized
+        if self._state_manager is None:
+            self._state_manager = InMemoryStateManager(DictState())
```
masci: why don't we use the init method here?

src/workflows/context/context.py
```diff
@@ -79,11 +81,35 @@ def __init__(
...
+                )
+
+            # State manager already initialised and compatible – nothing to do
+            return
+
+        # First-time initialisation
+        self._state_manager = InMemoryStateManager(state_class)
+
+    @property
+    def state(self) -> InMemoryStateManager[MODEL_T]:
```
masci: I would expect that `state` returns the state I specified to be `MyState` but I get a manager instead. A yellow flag this might be a problem is the `get_all` and `set_all` methods, where it's not clear what `all` is.
I haven't tried but I think we have two options to align semantic here:
1. We make the existence of the manager explicit: this proporty becomes `state_manager` and setter/getter become a writable property `ctx.state_manager.state`.
2. We remove the concept of the manager, moving its code to a base `State` class that `MyState` must inherit from. This way this property could return the actual `MODEL_T` state

masci (commented):
Left a comment about an indirection level it would be nice to remove if possible

src/workflows/context/context.py
logan-markewich: the init method is async, and we can't have an async `@property` 👀

logan-markewich: special case since we don't need to check the existing class (there is none)

src/workflows/context/context.py
logan-markewich: for 1. `ctx.state_manager.state` can't work unless we make it async. `get_state()` ?

oh interesting idea on 2 🤔 But It also removes the general idea of async. The reason I've made it async is I can imagine a future where the state for a workflow may live in some db (and the state could be so large that you might avoid loading the entire thing and use the get/set methods instead)

If that doesn't sound valuable I can just remove all async and go the 2 route

masci: It looks like the blocker in any direction is having a property, do we really need it? I'm fine with something like `ctx.get_state()`

logan-markewich: I feel like the property was a convenient way to expose all the state-related methods without polluting the `Context` API further 🤔 

What about this
```python
await ctx.store.get_state()
await ctx.store.set_state(NewStateCLS())
await ctx.store.get("key")
await ctx.store.set("key", "val")
```

logan-markewich: I've pushed a rename similar to the above, let me know what you think -- I think its much nicer now

masci: What about (no code change, only renaming)
```py
await ctx.state.get()
await ctx.state.set(NewStateCLS())
await ctx.state.get_item("key")
await ctx.state.set_item("key", "val")
```

logan-markewich: @masci that looks alright, but I guess its a matter of how similar we want a new `get`/`set` API to be compared to the old one (the current code makes any transition slightly less painless I think? But tbd if thats actually important)

Doesn't that also bring back the confusion that `.state` is returning a state manager and not the actual state?

masci: Arf you're right, sorry for the noise - respecting the return type is a big plus and `ctx.store` works! Overall I like this better than "manager", thanks for inluding my feedback 🎉

masci (approved):
🚀

PR #17, opened 2025-06-26
add example notebooks

logan-markewich:
- [x] Feature walkthrough
- [x] Agent workflow
- [ ] HITL LlamaExtract workflow


logan-markewich:
Going to add the HITL example in another PR

PR #21, opened 2025-07-02
Add async context manager to workflow context

logan-markewich:
Adds an async manager to ensure locked access to the workflow state

```
    state.val += 1
    await asyncio.sleep(0.01)
    if state.val % 2 == 0:
        state.val -= 1
```

Passes the state and sets the lock, saves the state, and then releases the lock

### DIFF
--- a/src/workflows/context/state_store.py
+++ b/src/workflows/context/state_store.py
@@ -177,7 +177,7 @@ class InMemoryStateStore(Generic[MODEL_T]):
         return cls(state_instance)  # type: ignore
 
     @asynccontextmanager
-    async def state(self) -> AsyncGenerator[MODEL_T, None]:
+    async def edit_state(self) -> AsyncGenerator[MODEL_T, None]:
         """
         A context manager for editing the state.
         The state will be locked while the context manager is active.
@@ -185,7 +185,7 @@ class InMemoryStateStore(Generic[MODEL_T]):
 
         Example:
         ```python
-        async with ctx.store.state() as state:
+        async with ctx.store.edit_state() as state:
             state.name = "John"
             state.age = 30
         ```
--- a/tests/test_workflow_typed_state.py
+++ b/tests/test_workflow_typed_state.py
@@ -70,7 +70,7 @@ class ParallelWorkflow(Workflow):
 
     @step
     async def worker(self, ctx: Context[SomeState], ev: WorkerEvent) -> ResultEvent:
-        async with ctx.store.state() as state:
+        async with ctx.store.edit_state() as state:
             state.val += 1
             await asyncio.sleep(0.01)
             if state.val % 2 == 0:


adrianlyjak:
@logan-markewich Do you have any example problems that this solving? Is it to have an atomic view of the state for the duration of the context, and have that actually be committed without concern over other updates?

One thing that's maybe weird is that it looks like if any other updates come in that don't use the state() lock, that happen concurrently while there's a coroutine that has the lock, those updates will be obliterated.

As far as naming, the concept of transaction kind of seems relevant here. Maybe calling it `.transaction()` would be helpful, and to clarify that it will always commit the full state at the end.

if committing everything is a bad idea, another idea would be to somehow have the state copy track what values were set, and only set those values, but that might also be complicated and also unintuitive

logan-markewich:
@adrianlyjak in terms of the problem this is solving, if you look at this notebook
https://github.com/run-llama/workflows-py/blob/main/examples/document_processing.ipynb

You'll see you always need symmetrical get/set calls. This is problematic because 
1. Its annoying and easy to forget to call set again
2. Is not behind a lock, so other steps can modify at runtime and you'll lose data. Notice that `get_state()` returns a copy of the state.

There is no write access that is unprotected by a lock, fyi 

Re naming, `async with ctx.store.transaction() as state` maybe still doesn't feel right? But eh?

adrianlyjak:
@logan-markewich aha, that makes sense. So normal updates would be blocked by the lock still, and happen subsequently. Agree that the transaction name still doesn't read great, but I do sort of like advertising the concept as a transaction, makes it more intuitive what it does, at least to me. Maybe that just belongs in docs though

Some more bad ideas to pick from if you see anything you like 😄 (I'm fine with `.state()`


```py
async with ctx.store.state_transact() as state:
    pass
```

```py
async with ctx.store.locked_state() as state:
    pass
```

```py
async with ctx.store.state() as state_transact: # maybe more of how we document it
    pass
```

adrianlyjak (approved):

src/workflows/context/state_store.py
```diff
@@ -175,6 +176,25 @@ def from_dict(
 
         return cls(state_instance)  # type: ignore
 
+    @asynccontextmanager
+    async def state(self) -> AsyncGenerator[MODEL_T, None]:
+        """
+        Context manager to get a copy of the state.
```
adrianlyjak: this seems like it could be documented a bit more. Maybe like

```
Context manager to get a copy of the state within a lock. No other consumers can update the state
while this lock is held. Useful for transactional processes where consistency of reads and writes are
needed over the process of the process. State is updated to the final value when the context manager
is released.
```

logan-markewich: Yea I can write a better docstring here 👍🏻

tests/test_workflow_typed_state.py
```diff
@@ -38,3 +40,61 @@ async def test_typed_state() -> None:
...
+
+
+@pytest.mark.asyncio
+async def test_typed_state_with_context_manager() -> None:
+    wf = ParallelWorkflow()
+
+    result = await wf.run()
+
+    # Should only be 1 since the context manager locks the state
+    assert result == 1
```
AstraBert, line 100: Maybe I did not understood it correctly, but shouldn't this always be 1?
I mean, if every time that the value is even, you subtract 1, you virtually never reach a value that is different from 1? (even if there was not a lock(?))

AstraBert (approved):
Looks good to me, just the thing about the test which is not 100% clear and the documenting (as per Adrian's comment)

tests/test_workflow_typed_state.py
adrianlyjak: its a concurrency thing If you think of unrolling the concurrent operations, it looks more like this

```
+1
+1
+1
+1
...
asyncio.sleep()
asyncio.sleep()
...
if state % 2 == 0
  -1
if state % 2 == 0
...
```

logan-markewich: If you replace the context manager with normal get_state/set_state calls, you will get something like 3. This is because other asyncio tasks are checking out, editing, and commiting state changes at the same time.

(I confirmed this by trying it myself locally lol)

By using the context manager, it ensures sequential operations on the state

AstraBert: Oh yeah ok, it makes 100% sense now... I didn't think of it in terms of concurrency, thanks for clarifying! :))

PR #22, opened 2025-07-03
Adding an observability example

AstraBert:
As per title :)

### DIFF
--- a/examples/workflows_observability_pt1.ipynb
+++ b/examples/workflows_observability_pt1.ipynb
@@ -63,20 +63,20 @@
      "base_uri": "https://localhost:8080/"
     },
     "id": "9yQ8Z5zByLuj",
-    "outputId": "667f07b7-1cfb-45fb-e4f5-624d177fb8dc"
+    "outputId": "274bc1da-8d3f-4947-b7a4-fdec0276084b"
    },
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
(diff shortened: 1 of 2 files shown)


logan-markewich (commented):
This is great! A ton of good content in here

I might actually split this up into two notebooks given the amount of content covered?

1. One to explain how instrumentation works. With a super basic workflow that has one extra helper function thats instrumented, and emits one custom event
2. Another workflow that highlights observability with your custom email example (this notebook might assume that you've went through the other one first? Or at least link to it)

Thoughts?

AstraBert:
<p>Yeah, because the console exporter would pollute the output with too much logs, so we can use a file exporter and print it out when we need. Yes, I will also add the fact that you can use other exporters, such as Jaeger or Prometheus :)</p>

---

 <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-3034254865'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

AstraBert:
<p>Yeah, I'll move it up and reword it a bit!</p>

---

 <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-3034254908'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

AstraBert:
<p>Yeah, the problem with this is and the above is that it is JSON formatted, but is a sort of JSONL that is very difficult to parse, I will try to make it exported in such a way that it is easier to parse</p>

---

 <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-3034254893'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

AstraBert:
<p>Oh, I didn't know about this, thank you so much!</p>

---

 <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-3034254878'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

AstraBert:
Yeah, seems good to me! I'll be working on splitting them and get them out later today :))

logan-markewich:
<p>Intersting -- yea even if we read the file as a string and only print the first 50 lines or something, its just a lot to even scroll past when you have the notebook open</p>

---

 <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-3034254893'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

examples/workflows_observability_pt1.ipynb
```diff
@@ -0,0 +1,993 @@
+{
```
logan-markewich, line 1: <p>Shouldn't we just use the custom event and span above to demonstrate the events/spans working, rather than jumping to openai?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-2186079247'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

logan-markewich, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.        @dispatcher.span
</pre><p>Technically, every step in a workflow is already instrumented</p><p> </p><p>I think rather than creating two of these <code>TextAnalysisWorkflows</code> , we can just use the function and custom event we defined in step 2 inside a workflow.</p><p> </p><p>This way, users will learn that</p><ol><li>They can instrument their own functions and dispatch their own events</li><li>If they use their custom stuff inside a workflow, since the workflow steps are already pre-instrumented, their custom spans and events get nested as expected </li></ol><p> </p><p>I think that will drive home the basics around how this is working </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-2186079248'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

examples/workflows_observability_pt2.ipynb
```diff
@@ -0,0 +1,1816 @@
+{
```
logan-markewich, line 1: <pre class="ql-syntax" spellcheck="false">Line #12.        @dispatcher.span
</pre><p>small nit: these should already be instrumented</p><p><a href="https://github.com/run-llama/workflows-py/blob/3a6d980e7bf6ed497ecf6d299b232bc85cbdcf09/src/workflows/context/context.py#L676" target="_blank">https://github.com/run-llama/workflows-py/blob/3a6d980e7bf6ed497ecf6d299b232bc85cbdcf09/src/workflows/context/context.py#L676</a></p><p> </p><p>If this isn't working though, let me know (thats an issue then lol)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/workflows-py/pull/22/discussion/#comment-2186079249'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

PR #28, opened 2025-07-22
feat: add WorkflowServer to serve workflows as web services

masci:
Serve workflows as web services with HTTP endpoints, real-time event streaming, and support for custom StartEvent objects.

WorkflowServer:
- exposes HTTP endpoints for workflow execution, based on Starlette
- can be integrated in existing ASGI applications (e.g. FastAPI) or run standalone using uvicorn under the hood
- supports both synchronous (/workflows/{name}/run) and asynchronous (/workflows/{name}/run-nowait) execution

Event Streaming:
- /events/{handler_id} endpoint for streaming workflow events in real-time
- supports both NDJSON and Server-Sent Events (SSE) formats

Custom StartEvent Support:
- Accepts custom `StartEvent` objects via `start_event` parameter containing the serialized event instance

Dependencies Added:
- starlette>=0.39.0
- uvicorn>=0.32.0
- httpx>=0.25.0 for the testing client (dev dependency)

## API Endpoints

| Method | Endpoint                     | Description                         |
|--------|------------------------------|-------------------------------------|
| GET    | /health                      | Health check                        |
| GET    | /workflows                   | List available workflows            |
| POST   | /workflows/{name}/run        | Execute workflow synchronously      |
| POST   | /workflows/{name}/run-nowait | Execute workflow asynchronously     |
| GET    | /results/{handler_id}        | Get async workflow result           |
| GET    | /events/{handler_id}         | Stream workflow events (NDJSON/SSE) |

## Example Usage
```py
# Server setup
server = WorkflowServer()
server.add_workflow("greeting", GreetingWorkflow())
await server.serve(host="0.0.0.0", port=8000)

# Client usage with custom StartEvent
curl -X POST http://localhost:8000/workflows/greeting/run \
  -H "Content-Type: application/json" \
  -d '{"start_event": "{\"__is_pydantic\": true, \"value\": {\"_data\": {\"name\": \"Alice\"}}, \"qualified_name\": \"workflows.events.StartEvent\"}"}'

# Event streaming
curl http://localhost:8000/events/{handler_id}?sse=true
```


src/workflows/server/server.py
```diff
@@ -0,0 +1,161 @@
...
+            ),
+            Route(
+                "/results/{handler_id}",
+                self._get_workflow_result,
+                methods=["GET"],
+            ),
+            Route(
+                "/health",
+                self._health_check,
+                methods=["GET"],
```
adrianlyjak, line 74: Thoughts about tracking historical runs here, how persistence works? I think this is useful as a small thing to get up and running, but I'm wondering about more complex needs for llama deploy and whether this will be usable.

src/workflows/server/server.py
```diff
@@ -0,0 +1,161 @@
...
+
+        try:
+            body = await request.json()
+            context_data = body.get("context")
+            run_kwargs = body.get("kwargs", {})
+
+            context = None
+            if context_data:
+                context = Context.from_dict(workflow=workflow, data=context_data)
+            result = await workflow.run(context=context, **run_kwargs)
```
adrianlyjak: This can't really capture typed start events. This would be passing a dict in place of the start event

src/workflows/server/server.py
```diff
@@ -0,0 +1,161 @@
...
+                self._run_workflow,
+                methods=["POST"],
+            ),
+            Route(
+                "/workflows/{name}/run-nowait",
+                self._run_workflow_nowait,
+                methods=["POST"],
+            ),
+            Route(
+                "/results/{handler_id}",
```
adrianlyjak, line 62: Are you planning on returning the streamed events? 

Would that capture input and output events? (Useful to model conversational turns with input required events).

This also relates to the start_event serialization/deserialization. A consumer likely needs some way of interpreting the event type. In JsonSerializer we've historically used the fully qualified python name (module + name), but it might be nice to have some mechanism to provide an event name.

adrianlyjak (commented):
added some misc questions to help me understand the direction this is going, and how it should fit with llama deploy

src/workflows/server/server.py
masci: The idea is to keep this as simple as possible, so at this moment there's no plan to introduce persistence of historical runs, the outer app would be responsible for that.

src/workflows/server/server.py
masci: I'm adding an endpoint for streaming events produced by the workflow, it'll look a lot like this initially https://github.com/run-llama/llama_deploy/blob/main/llama_deploy/apiserver/routers/deployments.py#L153-L158

src/workflows/server/server.py
```diff
@@ -0,0 +1,193 @@
...
+                context = Context.from_dict(workflow=workflow, data=context_data)
+            self._handlers[handler_id] = workflow.run(ctx=context, **run_kwargs)
+
+            return JSONResponse({"handler_id": handler_id, "status": "started"})
+        except Exception as e:
+            return JSONResponse({"error": str(e)}, status_code=500)
+
+    async def _get_workflow_result(self, request: Request) -> JSONResponse:
+        handler_id = request.path_params["handler_id"]
+        handler = self._handlers.pop(handler_id, None)
```
adrianlyjak: What about the handlers GC themselves after some configurable TTL? The result only being retrievable once is a little semantically surprising--generally `GET`s are supposed to be idempotent

masci: I wanted to replicate the behaviour in the Python API (you can only await the result once) but I see how it doesn't fit the REST paradigm... I have to rework this method anyways (right now it just awaits the result, possibly timing out on the client side and losing the result forver) I'll add results caching

src/workflows/server/server.py
```diff
@@ -0,0 +1,220 @@
...
+                self._run_workflow_nowait,
+                methods=["POST"],
+            ),
+            Route(
+                "/results/{handler_id}",
+                self._get_workflow_result,
+                methods=["GET"],
+            ),
+            Route(
+                "/events/{handler_id}",
```
logan-markewich, line 67: This type of endpoint basically means you can never replicate/scale this server right? (Since we are storing handlers/contexts in the sever, making it stateful)

Not saying that has to be a priority, but would be an easy foot gun for a person to miss

src/workflows/server/server.py
adrianlyjak: aha, I didn't realize the existing llama deploy API was mutable

src/workflows/server/server.py
```diff
@@ -0,0 +1,220 @@
...
+        # Get raw_event query parameter
+        sse = request.query_params.get("sse", "false").lower() == "true"
+        media_type = "text/event-stream" if sse else "application/x-ndjson"
+
+        async def event_stream(handler: WorkflowHandler) -> AsyncGenerator[str, None]:
+            serializer = JsonSerializer()
+
+            async for event in handler.stream_events():
+                serialized_event = serializer.serialize(event)
+                if sse:
```
adrianlyjak, line 174: ❤️

src/workflows/server/server.py
```diff
@@ -0,0 +1,220 @@
...
+        if "name" not in request.path_params:
+            raise HTTPException(detail="'name' parameter missing", status_code=400)
+        name = request.path_params["name"]
+
+        if name not in self._workflows:
+            raise HTTPException(detail="Workflow not found", status_code=404)
+
+        return self._workflows[name]
+
+    async def _extract_run_params(self, request: Request, workflow: Workflow) -> tuple:
```
adrianlyjak, line 199: Thoughts about adding some sort of plugin architecture that mostly gets injected here? This looks like a nice choke point for doing things like:

- look for some extra request parameters or body to extend the API (e.g. mainly thinking about the use case of loading an existing context from a data store)
- Could add a checkpoint_callback for context persistence (to store said contexts)
- Could generally track new runs, and put them in a datastore

Also thinking about these
```
        self._workflows: dict[str, Workflow] = {}
        self._contexts: dict[str, Context] = {}
        self._handlers: dict[str, WorkflowHandler] = {}
```

Maybe a plugin isn't the best concept, but more like a pluggable "backend". Ideally you'd be able to have a simple layer here to run this in memory as a single replica like it is, but for other use cases, still be able to share this api, but drop in a more full featured backend

tests/server/test_server_endpoints.py
```diff
@@ -0,0 +1,408 @@
...
+        # Stream events
+        response = await client.get(f"/events/{handler_id}")
+        assert response.status_code == 200
+        assert response.headers["content-type"] == "application/x-ndjson"
+
+        # Collect streamed events
+        events = []
+        async for line in response.aiter_lines():
+            if line.strip():
+                import json
```
adrianlyjak: nit: a lot of this inline json imports

tests/server/test_server_endpoints.py
```diff
@@ -0,0 +1,408 @@
...
+        assert response.headers["content-type"].startswith("text/event-stream")
+
+        # Collect streamed events
+        events = []
+        current_event = {}
+        async for line in response.aiter_lines():
+            line = line.strip()
+            if line.startswith("event: "):
+                # Extract event type
+                current_event["event_type"] = line[7:]  # Remove "event: " prefix
```
adrianlyjak: nit: 
```suggestion
                current_event["event_type"].removeprefix("event: ")
```

src/workflows/server/server.py
```diff
@@ -0,0 +1,220 @@
...
+            async for event in handler.stream_events():
+                serialized_event = serializer.serialize(event)
+                if sse:
+                    # need to convert back to str to use SSE
+                    event_dict = json.loads(serialized_event)
+                    yield f"event: {event_dict.get('qualified_name')}\ndata: {json.dumps(event_dict.get('value'))}\n"
+                else:
+                    yield f"{serialized_event}\n"
+
+                await asyncio.sleep(0.01)
```
adrianlyjak: nit: I think sleeping 0 should usually be enough to allow the next tasks in line in the event loop to run
```suggestion
                await asyncio.sleep(0)
```

src/workflows/server/server.py
adrianlyjak: maybe related: https://github.com/run-llama/workflows-py/pull/28#discussion_r2228651533

masci: Correct, if we want to support "run-no-wait" the server has to be stateful. We can provide a way to share the server state among server instances in a subsequent PR, but this will likely involve some database instance that's accessible from all the instances.

src/workflows/server/server.py
masci: Agree, similar to https://github.com/run-llama/workflows-py/pull/28#discussion_r2228722929 we should make the server state pluggable so it can be shared

pyproject.toml
```diff
@@ -20,7 +21,9 @@ requires-python = ">=3.9"
 dependencies = [
   "eval-type-backport>=0.2.2 ; python_full_version < '3.10'",
   "llama-index-instrumentation>=0.1.0",
-  "pydantic>=2.11.5"
+  "pydantic>=2.11.5",
+  "starlette>=0.39.0",
+  "uvicorn>=0.32.0"
```
logan-markewich, line 30: do we want to hardcode these dependencies? Or have them be extras?

logan-markewich: (just tested and this adds ~5MB to the package size, from 10MB to 15MB)

masci: ah yes, this should go in a `server` extra or something like that (if you have suggestions for the name of the extra please let me know!)

logan-markewich: `pip install llama-index-workflows[server]` is fine by me 👍🏻

examples/server/fastapi_server_example.ipynb
logan-markewich, line 1: Nit: missing a title with the `# abc` formatting

examples/server/server_example.ipynb
logan-markewich, line 1: Is this a duplicate?

pyproject.toml
```diff
@@ -7,7 +7,8 @@ dev = [
   "pre-commit>=4.2.0",
   "pytest>=8.4.0",
   "pytest-asyncio>=1.0.0",
-  "pytest-cov>=6.1.1"
+  "pytest-cov>=6.1.1",
+  "httpx>=0.25.0"
```
logan-markewich, line 11: Does httpx actually need a min version?

pyproject.toml
```diff
@@ -23,6 +24,12 @@ dependencies = [
   "pydantic>=2.11.5"
 ]
 
+[project.optional-dependencies]
+server = [
+  "starlette>=0.39.0",
+  "uvicorn>=0.32.0"
```
logan-markewich, line 30: Similar question here, is the min version needed in this case?

src/workflows/server/server.py
```diff
@@ -0,0 +1,231 @@
...
+            raise HTTPException(detail="Workflow not found", status_code=404)
+
+        return self._workflows[name]
+
+    async def _extract_run_params(self, request: Request, workflow: Workflow) -> tuple:
+        try:
+            body = await request.json()
+            context_data = body.get("context")
+            run_kwargs = body.get("kwargs", {})
+            start_event_str = body.get("start_event")
```
logan-markewich, line 204: Should we type the request object for thos endpoint or nah?

pyproject.toml
masci: Not sure, it's just what `uv add` does automatically, guess it guarantees a baseline at the moment this new code is added

examples/server/server_example.ipynb
masci: This shows how to use a vanilla `WorfklowServer` without a higher level framework like FastAPI

src/workflows/server/server.py
masci: We could, but we need to keep it under control to avoid rewriting FastAPI

pyproject.toml
logan-markewich: llama-deploy is setting `uvicorn[standard]>=0.12.0` fyi

pyproject.toml
logan-markewich: I think this is fine for now, just checked llama-index-core

Issue #27, opened 2025-07-21
How to clear context?


albertauyeung:
In https://github.com/run-llama/workflows-py/blob/702c0ba2ec4c8a0bf41edce2127c386ab7456ed4/src/workflows/context/context.py#L552 it is mentioned that
```
DEPRECATED: Use `await ctx.store.set(StateCLS())` instead.
This method is deprecated and will be removed in a future version.
```

Anyone know how do we use the new statement instead of calling `ctx.clear()`?

What should we put in the place of `StateCLS()`? An example would be appreciated. Thank you!

logan-markewich:
@albertauyeung you have two options

1. Using a typed state
```python
class MyState(BaseModel):
  val: int = Field(default=0)

class MyWorkflow(Workflow):
  @step
  async def my_step(self, ctx: Context[MyState], ev: StartEvent) -> StopEvent:
    return StopEvent()

w = MyWorkflow()
ctx = Context(w)

# Reset the state
w.store.set_state(MyState())
```

2. Using the default state

```python
from workflows.context.state_store import DictState

class MyWorkflow(Workflow):
  @step
  async def my_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
    return StopEvent()

w = MyWorkflow()
ctx = Context(w)

# Reset the state
w.store.set_state(DictState())
```

I'm realizing now that the deprecation message is slightly out of date, I'll fix that. Should be `.set_state()`

albertauyeung:
Thanks for the reply! Would be great if you can add these to the docstring.

PR #30, opened 2025-07-23
properly move and support clear method

logan-markewich:
Fixes https://github.com/run-llama/workflows-py/issues/27

We deprecated the old `clear()` method, but
a) provided no (easy) alternative
b) had incorrect docstrings and deprecation warnings pointing to the wrong methods

I also updated some errors slightly, to better reflect issues when your state cls does not have default fields (I ran into this when writing the test)

### DIFF
--- a/src/workflows/context/state_store.py
+++ b/src/workflows/context/state_store.py
@@ -1,7 +1,7 @@
 import asyncio
 import warnings
 from contextlib import asynccontextmanager
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 from typing import Any, AsyncGenerator, Generic, Optional, TypeVar
 
 from workflows.events import Event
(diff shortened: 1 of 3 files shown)


src/workflows/context/state_store.py
```diff
@@ -245,6 +245,14 @@ async def set(self, path: str, value: Any) -> None:
             # Assign the final value
             self._assign_step(current, segments[-1], value)
 
+    async def clear(self) -> None:
+        """Clear the state."""
+        async with self._lock:
```
masci: I would call set_state() here so we have this logic in one place only

src/workflows/context/state_store.py
```diff
@@ -245,6 +245,14 @@ async def set(self, path: str, value: Any) -> None:
...
+            try:
+                self._state = self._state.__class__()
```
masci: Not sure if viable, but we could make a copy of the intial state in the init, or the first time the state changes, and restore it in this method. We could even change the name to reset() or restore() to better convey what happens here. This way we wouldn't need to enforce default values for all params.

adrianlyjak (approved):
LGTM

masci (changes requested):

src/workflows/context/state_store.py
logan-markewich: ah good catch

src/workflows/context/state_store.py
logan-markewich: We still need to enforce defaults due to https://github.com/run-llama/workflows-py/blob/4d4fecf31c9dd8870b4db5103990f32792886dff/src/workflows/context/context.py#L651

logan-markewich: (Users are setting `Context[MyState]` and we are assuming we can instantiate on the fly, I haven't thought about alternative UX's to exposing the state manager yet)

PR #35, opened 2025-07-29
chore: fix typing issues

masci:
- Fixed typing issues as reported by pyright, codebase is clean now.
- Extracted a dict-like base class for a better semantic of `DictState` that used to derive from `Event` only to use its dict-like behaviour


logan-markewich (approved):
Seems safe!

PR #38, opened 2025-07-30
feat: add startup script to ease server startup in Docker

masci:
Added a simple script to start a `WorkflowServer` instance by just passing the path to a script. This is mostly helpful in docker containers, we might add more features later and at some point we might turn it into a proper CLI.

Also added an example showing how to run a workflow server in a docker container.

### DIFF
--- a/examples/docker/requirements.txt
+++ b/examples/docker/requirements.txt
@@ -1,1 +1,1 @@
-git+https://github.com/run-llama/workflows-py.git@massi/docker#egg=llama-index-workflows[server]
+llama-index-workflows[server]


examples/docker/requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/run-llama/workflows-py.git@massi/docker#egg=llama-index-workflows[server]
```
masci: To be changed before merging

examples/docker/app.py
```diff
@@ -0,0 +1,33 @@
...
+    async def greet(self, ctx: Context, ev: StartEvent) -> StopEvent:
+        for i in range(3):
+            ctx.write_event_to_stream(StreamEvent(sequence=i))
+            await asyncio.sleep(0.3)
+
+        name = getattr(ev, "name", "World")
+        return StopEvent(result=f"Hello, {name}!")
+
+
+server = WorkflowServer()
```
logan-markewich, line 24: Should we note that calling it `server` is required?

masci: Yeah this should be extensively documented... let me see if we can be smart here and inspect type

masci: Removed the need of calling the server instance `server`

PR #44, opened 2025-08-04
chore: deprecate `ServiceManager` and `add_workflows()`

### LINKED ISSUE #42: [Proposal] Deprecate subworkflows
Subworkflows have no advantage compared to instantiating a new workflow in a step. As we work on a better UX (see #37), we should remove the `ServiceManager` and push users towards creating subworkflows in-line, possibly using a simplified api as stated in #37

masci:
Fixes #42

### DIFF
--- a/src/workflows/service.py
+++ b/src/workflows/service.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


src/workflows/service.py
```diff
@@ -19,9 +20,16 @@ class ServiceManager:
 
     A Service is nothing more than a workflow instance attached to another workflow.
     The service is made available to the steps of the main workflow.
+
+    DEPRECATED: The ServiceManager class is deprecated and will be removed in a future version.
     """
 
     def __init__(self) -> None:
+        warnings.warn(
+            "ServiceManager is deprecated and will be removed in a future version.",
```
logan-markewich, line 29: ooc should we point users to alternatives? Or nah?

masci: Not sure. Consider that "Service" is never used in the docs or the api, and this was powering a feature called "sub-workflows" so I think it's better to just remove it and pretend it never happened 😅.

PR #45, opened 2025-08-04
chore: deprecate old workflow.send_event() method

masci:
Method has never been officially deprecated, but it's been spitting a warning since a long time, let's just make it official so we can schedule it for removal.

### DIFF
--- a/src/workflows/workflow.py
+++ b/src/workflows/workflow.py
@@ -312,8 +312,7 @@ class Workflow(metaclass=WorkflowMeta):
         DEPRECATED: This method is deprecated and will be removed in a future version.
         """
         warnings.warn(
-            "send_event() is deprecated and will be removed in a future version. "
-            "There are no alternatives.",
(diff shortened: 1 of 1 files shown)


src/workflows/workflow.py
```diff
@@ -306,6 +306,18 @@ def _start(
...
 
     def send_event(self, message: Event, step: str | None = None) -> None:
+        """
+        Send an event to the workflow.
+
+        DEPRECATED: This method is deprecated and will be removed in a future version.
+        """
+        warnings.warn(
+            "send_event() is deprecated and will be removed in a future version. "
+            "There are no alternatives.",
```
logan-markewich: lol isn't `ctx.send_event()` the alternative?

Something like "Use `ctx.send_event()` on the Context object in the workflow step."

masci: That's stupid Claude, shame on me for not catching that in the diff

PR #50, opened 2025-08-06
fix: raise an error when streamed events have already been consumed

### LINKED ISSUE #49: Consuming streaming events twice hangs indefinitely
Example pseudo-code:

```py
workflow = MyWorkflow()
handle = workflow.run()

# This works as expected, loop is exited when StopEvent is sent
async for event in handle.stream_events():
    print("got event from first handle iteration", event)

# This will hang forever
async for event in handle.stream_events():
    print("got event from second handle iteration", event)
```

masci:
Fixes #49 

Initially I tried to check for the length of the queue along with the `ctx.is_running` flag, but this was dependent on the status of the async loop, making the check unreliable for simple workflows where consuming all the events was faster than shutting down the workflow runtime properly.

I then added a flag to the handler instance to be set the first (and last, at this point) time the StopEvent is streamed.


src/workflows/handler.py
```diff
@@ -36,15 +37,60 @@ def is_done(self) -> bool:
         return self.done()
 
     async def stream_events(self) -> AsyncGenerator[Event, None]:
+        """
```
logan-markewich, line 40: docstrings!!

src/workflows/handler.py
```diff
@@ -36,15 +37,60 @@ def is_done(self) -> bool:
...
+            msg = "All the streamed events have already been consumed."
+            raise WorkflowRuntimeError(msg)
+
         while True:
             ev = await self.ctx.streaming_queue.get()
 
             yield ev
 
             if isinstance(ev, StopEvent):
+                self._all_events_consumed = True
```
logan-markewich, line 93: Just confirming my sanity: There's no way the handler ever gets reused (unlike the context) so we don't need a mechanism to reset this back to false, right?

masci: For now no, if anybody is doing this they're currently getting stuck waiting forever on the queue. We can revisit this if we want to support streaming events mutliple times.

PR #58, opened 2025-08-07
Remove stepwise execution

masci:


src/workflows/workflow.py
```diff
@@ -364,10 +349,6 @@ async def _run_workflow() -> None:
         asyncio.create_task(_run_workflow())
         return result
 
-    def is_done(self) -> bool:
-        """Checks if the workflow is done."""
-        return self._stepwise_context is None
-
```
masci, line 370: This was only relevant within a stepwise execution, but I see how the name can be misleading and we used it wrong ourselves in the tests, it's better to 🔥 it

logan-markewich:
This reminds me that we will have to do a sweep through the docs before 2.0

PR #60, opened 2025-08-07
Migrate docs from llama_index to workflows-py repo

masci:


docs/docs/_static/js/mendablesearch.js
logan-markewich, line 1: Oh this is old, we dont use mendable anymore

docs/docs/javascript/llms_example.js
logan-markewich, line 1: I would delete this too

docs/docs/javascript/mendablesearch.js
logan-markewich, line 1: This too

logan-markewich (approved):

PR #64, opened 2025-08-11
feat: simplify the fan-in/fan-out logic

### LINKED ISSUE #36: [Feature Request] Improve the map-reduce (fan-out/gather) syntax
Currently, to do a map reduce pattern, you need a decent amount of boilerplate

```py
import random

from workflows.events import Event, StartEvent, StopEvent


class TaskRequested(Event):
    pass


class TaskReady(Event):
    pass

class FanOutFanIn(Workflow):
    @step
    async def fan_out(self, ctx: Context, ev: StartEvent) -> TaskRequested | None:
        times = random.randint(1, 10)
        await ctx.store.set("times", times)
        for _ in range(times):
            ctx.send_event(TaskRequested())

    @step
    async def process_task(self, ev: TaskRequested) -> TaskReady:
        # do work
        return TaskReady()

    @step
    async def fan_in(self, ctx: Context, ev: TaskReady) -> StopEvent | None:
        num_to_collect = await ctx.store.get("times")
        events = ctx.collect_events(ev, [TaskReady] * num_to_collect)
        if events is not None
            return StopEvent(results=events)
```

Notice several things
- extremely verbose, lots of places to make a mistake
- technically incorrect step annotations to make the workflow validator happy (the map step doesn't actually "return" MapEvent)
- needing to know ahead of time how many events to collect

I think this pattern really starts to break down when you get more complex patterns. What if I have several levels of mapping?

I think ideally, there is a better way to do this, or at the very least, nicer syntactic sugar around this

AstraBert:
Closes #36 

With the introduction of the `dispatch` and `receive` methods, we simplify the logic behind  the current fan-in/fan-out (or map-reduce) approach, as detailed in the referenced issue


src/workflows/context/context.py
```diff
@@ -375,6 +376,27 @@ def send_event(self, message: Event, step: str | None = None) -> None:
 
         self._broker_log.append(message)
 
+    async def dispatch(self, events: list[Event]) -> None:
```
masci: Do we need two methods with very different names to do pretty much the same thing? I would propose one of the following:
- Remove `send_event` and only rely on `dispatch`. Single events would be sent with `ctx.dispatch([ev])`. Alternatively, make `dispatch` take `Union[Event, Sequence[Event]]`. The overhead of keeping `dispatched_events` is negligible
- Rename this function to `send_events` so it's clear what it does compared to the existing `send_event`

src/workflows/context/context.py
```diff
@@ -375,6 +376,27 @@ def send_event(self, message: Event, step: str | None = None) -> None:
...
+        dispatched = cast(dict, await self.store.get("dispatched_events", default={}))
+
+        for event in events:
+            if type(event) in dispatched:
+                dispatched[type(event)] += 1
+            else:
+                dispatched.update({type(event): 1})
```
masci: You can get rid of this using a `defaultdict`

src/workflows/context/context.py
```diff
@@ -375,6 +376,27 @@ def send_event(self, message: Event, step: str | None = None) -> None:
...
+            self.send_event(event)
+
+        await self.store.set("dispatched_events", dispatched)
```
masci: You can probably use the `with self.store.edit_state()` paradigm instead

src/workflows/context/context.py
```diff
@@ -375,6 +376,27 @@ def send_event(self, message: Event, step: str | None = None) -> None:
...
+
+    async def receive(
+        self, event: Event, event_type: Type[Event]
+    ) -> Union[list[Event], None]:
+        dispatched = cast(dict, await self.store.get("dispatched_events", default={}))
+        to_collect = dispatched.get(event_type, None)
```
masci:
```suggestion
        to_collect = dispatched.get(event_type)
```

masci (changes requested):

src/workflows/context/context.py
logan-markewich: The issue with renaming to send_event is this fn is async, the other is sync 🤔

AstraBert: If we make this part of v2, we can introduce the breaking change with the renaming, although I'm not super fond of it

AstraBert: Also, if we remove send_event this would not work since it is basically a wrapper around it... I am, tho, in favor of keeping send_event and renaming this to send_events

src/workflows/context/context.py
AstraBert: Mmhhh but the state doesn't have the `dispatched_events` attribute, so how can I set it?

logan-markewich: so, the default state is dictstate, you can set anything there magically :) 

But this actually raises another issue with this approach -- if the user sets a custom state model, this will actually fail unless their state model has dispatched_events in it

logan-markewich: We might need a solution that doesn't abuse the state store or we need some workaround.... (I'm realizing now that wait_for_event has the same bug 😬)

AstraBert: A job queue as @masci was suggesting?

AstraBert: We add it as a private attribute and we do not touch the state store when storing/fetching events

docs/docs/understanding/workflows/concurrent_execution.md
```diff
@@ -24,6 +24,29 @@ class ParallelFlow(Workflow):
...
+    async def start(self, ctx: Context, ev: StartEvent) -> StepTwoEvent | None:
+        events = [
+            StepTwoEvent(query="Query 1"),
+            StepTwoEvent(query="Query 2")
+            StepTwoEvent(query="Query 3")
+        ]
+        await ctx.dispatch(events)
+
+    @step(num_workers=4)
+    async def step_two(self, ctx: Context, ev: StepTwoEvent) -> StopEvent:
```
adrianlyjak: won't this stop the workflow after the first event since this returns a `StopEvent`? Might be better to leave it open ended/incomplete rather than returning a stop event

docs/docs/understanding/workflows/concurrent_execution.md
```diff
@@ -62,6 +85,42 @@ The `step_three` step is fired every time a `StepThreeEvent` is received, but `c
...
+        print("Running query ", ev.query)
+        await asyncio.sleep(random.randint(1, 5))
+        return StepThreeEvent(result=ev.query)
+
+    @step
+    async def step_three(
+        self, ctx: Context, ev: StepThreeEvent
+    ) -> StopEvent | None:
+        # wait until we receive 3 events
+        result = await ctx.receive(ev, type(ev))
```
adrianlyjak: nit: I'm not sure `receive` is the most intuitive name here for accumulating multiple events. What about `gather`?

Also, this might be too magical, but the bailing with `None` is kind of confusing. We could perhaps consider throwing a special exception from this call that the workflow infrastructure interprets the same as `None`. Like a `NotReady` exception or something

src/workflows/context/context.py
adrianlyjak: you'll need to make sure it serializes with the context still though

src/workflows/context/context.py
adrianlyjak: `send_events` seems like a reasonable solution, although the state mutation means that this is indeed a little special, so it's perhaps confusing to name them too closely. (the state mutation is also the cause of the async)

Another thought: I'm also perhaps somewhat concerned that this doesn't have a great continuity for how to fan out multiple times before joining. That's probably a rare scenario, just would be nice if we had a more composable pattern.

Like, I'm almost wondering if there's something we could do at the type level instead of checking for the receives? Probably not... it's maybe too dynamic (and this would be much more complicated).

But like, if you could somehow tag it as `events: All[SomeEventType, MaybeSomeOtherEventType]`, this could trigger the workflow step as pending as soon as one of the events is added to the queue, and then the step would only be run run once all of the selected types within the queue were resolved. And then `All` ends up being an annotated type for a list, so you'd have a list in the function.

Pie in the sky, but seemed worth mentioning.

src/workflows/context/context.py
logan-markewich: yea we have lots of handling for all the context attributes, we could add that to the list of things to handle

src/workflows/context/events_queue.py
```diff
@@ -0,0 +1,16 @@
+from collections import defaultdict
+
+from workflows.events import Event
+from typing import Sequence, Type, Dict, Any, Optional
+
+
+class EventsQueue:
```
masci: do you plan to add more responsibilities to this class? If not, I would discourage using a class to just wrap a dictionary usage

AstraBert:
So @masci I removed the EventsQueue class and renamed the methods: now they are called `send_events` and `receive_events`: I'm not super opinionated about the naming, I just think it makes sense to have the send-receive couple

masci:
@AstraBert I'm fine with `gather`, it's the two sets of methods doing basically the same thing that I think we should consolidate. I think having *only* `send_events/gather` is a good enough interface, a bit more annoying to use for single events but overall more obvious to use.

AstraBert:
Ok so let's get rid of send_event! :))

src/workflows/context/context.py
```diff
@@ -375,6 +384,30 @@ def send_event(self, message: Event, step: str | None = None) -> None:
...
+    def receive_events(
+        self, event: Event, event_type: Union[Type[Event], list[Type[Event]]]
+    ) -> Union[list[Event], None]:
+        ev_types: list[Type[Event]] = []
+        if isinstance(event_type, list):
+            for ev_type in event_type:
+                to_collect = cast(Optional[int], self._events_queue.get(ev_type))
+                if to_collect:
+                    ev_types.extend([ev_type] * to_collect)
+        else:
```
masci: Why supporting both a single event and a list of events? For simmetry with `send_events` I would only take a list

src/workflows/context/context.py
```diff
@@ -375,6 +384,30 @@ def send_event(self, message: Event, step: str | None = None) -> None:
...
+        for event in events:
+            self._send_event(event)
+
+    def receive_events(
+        self, event: Event, event_type: Union[Type[Event], list[Type[Event]]]
+    ) -> Union[list[Event], None]:
+        ev_types: list[Type[Event]] = []
+        if isinstance(event_type, list):
+            for ev_type in event_type:
+                to_collect = cast(Optional[int], self._events_queue.get(ev_type))
```
masci: You don't need this `cast` if you fully annotate the defaultdict like `self._events_queue: defaultdict[Type[Event], int] = defaultdict(int)`

src/workflows/context/context.py
```diff
@@ -375,6 +384,30 @@ def send_event(self, message: Event, step: str | None = None) -> None:
 
         self._broker_log.append(message)
 
+    def send_events(self, events: list[Event]) -> None:
```
masci: If we remove `send_event` we should support the `step` param here

src/workflows/context/context.py
```diff
@@ -375,6 +384,30 @@ def send_event(self, message: Event, step: str | None = None) -> None:
...
+        for event in events:
+            self._events_queue[str(type((event)))] += 1
+
+        for event in events:
```
masci, line 559: Let's merge these two for loops

src/workflows/context/context.py
```diff
@@ -384,26 +383,63 @@ def _send_event(self, message: Event, step: str | None = None) -> None:
 
         self._broker_log.append(message)
 
-    def send_events(self, events: list[Event]) -> None:
-        for event in events:
-            self._events_queue[str(type((event)))] += 1
+    def send_events(self, events: list[Union[Event, Tuple[Event, str]]]) -> None:
```
masci: Not sure I get it? I would expect a `step` param of type `str` here, like 
```py
def send_events(self, events: list[Event], step: str | None = None) -> None:
```

AstraBert: That would mean you send all the events to a single step, right?
The idea here is that you can pass a list of events (if you do not want to send them to a specific step) and/or tuples[Event, str] so that you have control on which event gets sent to which step

masci: I think that complicates things with little gain, you can still call `send_events` multiple times targeting different steps (not even sure what use case would benefit from targeting different steps)

src/workflows/context/context.py
```diff
@@ -535,6 +536,66 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> WorkerEvent | GatherEve
...
+            events (list[Event]): a list of Event objects to be sent
+            step (Optional[str]): a string representing the target step to which events should be sent. Defaults to None if not set.
+
+        Returns:
+            None
+
+        Examples:
+        ```
+        class MultipleEventsWorkflow(Workflow):
+        @step
```
logan-markewich: nit: class method needs indent

AstraBert:
Hey @masci @logan-markewich I added a draft of how we could add the fan-out/gather logic to the step decorator. I haven't modified tests yet since I wanted to get your opinion on the implementation logic first and understand if I am going in the right direction.

If you confirm, I'll work on it more and make it solid :))

src/workflows/context/context.py
```diff
@@ -712,6 +773,9 @@ async def _step_worker(
             if verbose and name != "_done":
                 print(f"Running step {name}")
 
+            if config.gather is not None:
+                self.receive_events(ev, config.gather)
```
logan-markewich: We want to skip running the step if receive_events returns none right?

src/workflows/context/context.py
```diff
@@ -480,17 +489,9 @@ async def synthesize(
...
-
-        If `step` is omitted, the event is broadcast to all step queues and
-        non-matching steps will ignore it. When `step` is provided, the target
-        step must accept the event type or a
-        [WorkflowRuntimeError][workflows.errors.WorkflowRuntimeError] is raised.
-
-        Args:
-            message (Event): The event to enqueue.
-            step (str | None): Optional step name to target.
+    def _send_event(self, message: Event, step: str | None = None) -> None:
```
masci, line 493: Let's keep this around as `send_event` and make it a shallow wrapper of `send_events`.

src/workflows/context/context.py
```diff
@@ -535,6 +536,66 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> WorkerEvent | GatherEve
...
+        ```
+        """
+        for event in events:
+            self._events_queue[str(type(event))] += 1
+            if not step:
+                self._send_event(event)
+            else:
+                self._send_event(event, step)
+
+    def receive_events(
```
masci: I see in the decorator draft you're thinking about a `gather` parameter - let's call this function `gather_events` so the relation will be obvious.

src/workflows/context/context.py
```diff
@@ -414,7 +423,7 @@ def _get_event_buffer_id(self, events: list[Type[Event]]) -> str:
         # Fall back to creating a stable identifier from expected events
         return ":".join(sorted(self._get_full_path(e_type) for e_type in events))
 
-    def collect_events(
```
masci, line 417: Let's keep this around and alias it to `gather_events` with a deprecation notice

masci (changes requested):
After the conversations we had offline I think it would be easier to keep `send_event` around and make it a wrapper of `send_events`, as well as making `collect_events` an alias for `gather_events`. This way we can make the upgrade path 1.x -> 2.0 a bit easier and avoid one deprecation warning in 1.3 (we still need a deprecation warning for `collect_events`).

src/workflows/context/context.py
```diff
@@ -535,6 +537,95 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> WorkerEvent | GatherEve
...
+        ev_types: list[Type[Event]] = []
+        for ev_type in event_types:
+            to_collect = self._events_queue.get(str(ev_type))
+            if to_collect:
+                ev_types.extend([ev_type] * to_collect)
+        if ev_types:
+            return self._collect_events(event, ev_types)
+        return None
+
+    collect_events = gather_events
```
masci: To keep `collect_events` backward compat we need a third param, `buffer_id: str | None = None`. You can go with a regular method with the old signature

AstraBert: Done! I also noticed  that `send_event` was not backward compatible so I also changed that

src/workflows/context/context.py
```diff
@@ -535,6 +537,121 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> WorkerEvent | GatherEve
...
+        @step
+        async def send_events(self, ev: InputEvent, ctx: Context):
+            ctx.send_events(events = [(OutputAEvent(), None), OutputBEvent()])
+        ## rest of the implementation
+        @step
+        async def collect_events(self, ev: ReceiveEvent, ctx: Context):
+            ctx.collect_events(ev, [OutputAEvent, OutputBEvent])
+        ```
+        """
+        return self.gather_events(ev, expected)
```
logan-markewich, line 657: if I'm understanding this correctly, `gather_events` has a different functionality compared to the old `collect_events()` API -- gather will check for all recorded events of a certain type and wait/collect them

the older `collect_events()` relies on the user to pass in how many to collect 

I think because if this we should call `self._collect_events()` here? What do you think?

AstraBert: Done!

masci: Correct me if I'm wrong, but I think `gather_events` can actually be a drop-in replacement for  `collect_events()`, while it's not the case the other way around.

This:
```py
events = ctx.collect_events(ev, [QueryEvent, RetrieveEvent])
# events is None if one of [QueryEvent, RetrieveEvent] wasn't sent
# events is [QueryEvent, RetrieveEvent] when both events are sent
```
is equivalent to this:
```py
events = ctx.gather_events(ev, [QueryEvent, RetrieveEvent])
# events is None if one of [QueryEvent, RetrieveEvent] wasn't sent
# events is [QueryEvent, RetrieveEvent] when both events are sent
```

The difference is `gather_events` can be used in a way that `collect_events` cannot. For example, if I have a step doing `send_events([QueryEvent, QueryEvent, QueryEvent])`, these are different:
```py
events = ctx.collect_events(ev, [QueryEvent])
# events returns [QueryEvent]

events = ctx.gather_events(ev, [QueryEvent])
# events returns [QueryEvent, QueryEvent, QueryEvent]
```

AstraBert: Yeah, and now that I am thinking of it, if `send_event` is now a simple wrapper around `send_events`, it might make sense to use `gather_events` within the `collect_event` wrapper.

masci (approved):

src/workflows/context/context.py
logan-markewich: I'm specifically thinking about the case where you do

```
ctx.collect_events(ev, [SomeEvent]*10)
```

I think the above won't work if we are just wrapping gather_events.... going to add a unit test to confirm

logan-markewich: ok we have tests for the above, and they work! (Although not sure how tbh lol)

logan-markewich (approved):

PR #70, opened 2025-08-18
feat: enable openapi export

adrianlyjak:
yaml doc strings are required to support exporting openapi docs from routes


masci (approved):
LGTM!

PR #73, opened 2025-08-21
feat: expose internal events

### LINKED ISSUE #31: [Feature Request] Expose a public API for observing the runtime
While the `Context()` currently manages a ton of stuff (asyncio queues, task workers for each workflow step, state, etc.), not all of this is exposed in a user-friendly way

It would be nice to have
- public access to the start event CLS
- public access to the stop event CLS
- public API for checking the runtime state (which steps are running, are all queues exhausted, what events are waiting to be processed?)
  - This could be implemented as hooks as well (when events get added to queues, when steps error, when the state store changes)

AstraBert:
Closes #31

In this PR, we try to expose (through stream_events) internal processes within the workflow (step marked as in progress, step running, state of queues...)

Optimally, we should also be able to capture and report the changes in the State

Example usage:

```python
from workflows.events import StartEvent, StopEvent, Event, InternalDispatchEvent

class SecondEvent(Event):
    greeting: str

class TestWf(Workflow):
    @step
    async def step_one(self, ev: StartEvent, ctx: Context) -> SecondEvent:
        print(ev.message)
        ctx.write_event_to_stream(ev)
        return SecondEvent(greeting="hello")
    @step
    async def step_two(self, ev: SecondEvent, ctx: Context) -> StopEvent:
        print(ev.greeting)
        ctx.write_event_to_stream(ev)
        return StopEvent(result="done")

async def main():
    wf = TestWf()
    handler = wf.run(message="hi")
    async for ev in handler.stream_events(expose_internal=True):
        if isinstance(ev, InternalDispatchEvent):
            print(f"Event {type(ev.data)}: {ev.data}")

    result = await handler
    print(result)

if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```
 Output:

```text
Event <class 'workflows.events._QueueStateEvent'>: queue_name='_done' queue_size=1
Event <class 'workflows.events._QueueStateEvent'>: queue_name='step_one' queue_size=1
Event <class 'workflows.events._QueueStateEvent'>: queue_name='step_two' queue_size=1
hi
hello
Event <class 'workflows.events._InProgressStepEvent'>: ev=StartEvent() name='step_one' in_progress=True
Event <class 'workflows.events._RunningStepEvent'>: name='step_one' running=True
Event <class 'workflows.events._RunningStepEvent'>: name='step_one' running=False
Event <class 'workflows.events._InProgressStepEvent'>: ev=StartEvent() name='step_one' in_progress=False
Event <class 'workflows.events._QueueStateEvent'>: queue_name='_done' queue_size=1
Event <class 'workflows.events._QueueStateEvent'>: queue_name='step_one' queue_size=1
Event <class 'workflows.events._QueueStateEvent'>: queue_name='step_two' queue_size=2
Event <class 'workflows.events._InProgressStepEvent'>: ev=SecondEvent(greeting='hello') name='step_two' in_progress=True
Event <class 'workflows.events._RunningStepEvent'>: name='step_two' running=True
Event <class 'workflows.events._RunningStepEvent'>: name='step_two' running=False
Event <class 'workflows.events._InProgressStepEvent'>: ev=SecondEvent(greeting='hello') name='step_two' in_progress=False
Event <class 'workflows.events._QueueStateEvent'>: queue_name='_done' queue_size=2
Event <class 'workflows.events._QueueStateEvent'>: queue_name='step_one' queue_size=1
Event <class 'workflows.events._QueueStateEvent'>: queue_name='step_two' queue_size=1
Event <class 'workflows.events._InProgressStepEvent'>: ev=StopEvent() name='_done' in_progress=True
Event <class 'workflows.events._RunningStepEvent'>: name='_done' running=True
done
```


src/workflows/events.py
```diff
@@ -239,4 +242,66 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+        wf = ExampleWorkflow()
+        handler = wf.run(message="Hello, who are you?")
+
+        async for ev in handler.stream_event(expose_internal=True):
+            if isinstance(ev, InternalDispatchEvent):
+                print(ev.data)
+        ```
+    """
+
+    data: Union[
```
logan-markewich: Ooc, what's the motivation for nesting everything in this class? Just to make the downstream isinstance check easier?

I might consider just using this as the base class for other internal events if thats the case 👀

AstraBert: Done!

src/workflows/events.py
```diff
@@ -239,4 +242,68 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+        async for ev in handler.stream_event(expose_internal=True):
+            if isinstance(ev, InternalDispatchEvent):
+                print(ev.data)
+        ```
+    """
+
+    pass
+
+
+class _InProgressStepEvent(InternalDispatchEvent):
```
logan-markewich: ooc, do we want to keep underscores for these class names? Technically these aren't private I think?

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+            if isinstance(ev, InternalDispatchEvent):
+                print(ev.data)
+        ```
+    """
+
+    pass
+
+
+class InProgressStepEvent(InternalDispatchEvent):
+    ev: Event = Field(description="Event related to the step progression")
```
masci: We agreed offline not to stream anything we don't control, and custom events can carry a lot of data, including non-serializeable types.

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+    name: str = Field(description="Name of the step")
+    in_progress: bool = Field(
```
masci: Why `in_progress` should be false in an event called `InProgressStepEvent`? The semantic is already determined by the event itsel, no?

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+        description="True when step is marked as in progress, False when step is removed from in progress steps."
+    )
+
+
+class RunningStepEvent(InternalDispatchEvent):
+    name: str = Field(description="Name of the step")
+    run_id: str = Field(description="Run ID of the step for better distinction")
+    running: bool = Field(
```
masci: Simila to the above, why `running` should be false in a `RunningStepEvent`?

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+        description="True when step is marked as running, False when step is removed from running steps."
+    )
+
+
+class StateModificationEvent(InternalDispatchEvent):
+    modification_type: Literal["updated_state"]
```
masci: What other types of modification do you envision?

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+
+
+class QueueStateEvent(InternalDispatchEvent):
+    queue_name: str
```
masci: nit: for symmetry with the other events, this should be just `name`

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+    queue_size: int
+
+
+class StepEmitEvent(InternalDispatchEvent):
```
masci: what's this for?

src/workflows/workflow.py
```diff
@@ -260,6 +260,7 @@ def _start(
                     config=step_config,
                     verbose=self._verbose,
                     run_id=run_id,
+                    step_run_id=str(uuid.uuid4()),
```
masci: If i get this right, I think `worker_id` is more clear?

src/workflows/context/context.py
```diff
@@ -749,12 +793,18 @@ async def _step_worker(
 
             # - check if its async or not
             # - if not async, run it in an executor
+            start_state = await self.store.get_state()
+            initial_state: dict | BaseModel = (
+                start_state._data.copy()
+                if isinstance(start_state, DictState)
+                else start_state.model_copy()
+            )
```
masci, line 803: This is pretty heavy, can we find a more efficient way? Like we could compare size (keys added or removed alter the state) and fallback to hashing the values (should be faster than copy)

src/workflows/context/context.py
```diff
@@ -795,6 +845,16 @@ async def _step_worker(
                 except Exception as e:
                     raise WorkflowRuntimeError(f"Error in step '{name}': {e!s}") from e
 
+            end_state = await self.store.get_state()
+            final_state: dict | BaseModel = (
+                end_state._data.copy()
```
masci: I think there's no need to copy the final state, we can just compare it

src/workflows/events.py
```diff
@@ -239,4 +240,53 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+    step_name: str
+    input_event: Event
+    output_event: Optional[Event]
```
masci: Same as `InProgressStepEvent`, we might end up streaming Megabytes of data because we don't control the events

src/workflows/events.py
AstraBert: in_progress = False -> The step has been removed from the queue of steps that are in progress. I kept it this way so that we do not need two different events to signal when a step is marked or removed from in progress.

src/workflows/events.py
AstraBert: Same logic as above :)

src/workflows/events.py
AstraBert: In my first design, I was thinking about `state_property_added`, `state_property_removed`, `state_property_updated` (for when we add, remove or update properties within the state). Optimally with some tweaks we would be able to detect and assign this modification types to this event :)

src/workflows/events.py
AstraBert: @logan-markewich suggested we added some logic for detecting what events go in and out of workflow steps, hence this event

src/workflows/context/context.py
AstraBert: Yup, makes sense, I'll work on it

src/workflows/events.py
AstraBert: Could it be a good idea to use the event type instead of the event itself? Or the stringified event type

src/workflows/events.py
AstraBert: As per comment below, maybe it would make sense to use the event type/stringified event type?

src/workflows/events.py
masci: I would change the name then, something like
```py
class StepStateChanged(Event):
    name: str
    event_name: str
    in_progress: bool
```
Note I would also remove the `Event` suffix across the board, is very noisy

src/workflows/events.py
masci: I'm not sure we should introduce such a level of detail before users ask for it, but I imagine the caller would want to know which property changed at that point, I think we can get away with:
```python
class StateChanged(Event):
    properties_added: list[str]
    properties_removed: list[str]
    properties_updated: list[str]
```

src/workflows/events.py
masci: Let's use `StepStateChanged` then? It already contains the event that's supposed to have triggered the state change, let's also add the output events there...

src/workflows/events.py
masci: We should use event names (or full path string) across the board

src/workflows/context/context.py
masci: Uhm now we have even more code and we still have the copy... I took a look but I don't see a robust way to make this work honestly. Even keeping the `copy`, that's shallow so we won't capture changes to mutable objects. We might explore a way to do this from the state class with a callback or something.

src/workflows/events.py
```diff
@@ -239,4 +241,90 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+        description="State of the step ('running', 'not_running', 'in_progress', 'not_in_progress', 'exited')"
+    )
+    worker_id: str = Field(description="ID of the worker that the step is running on")
+    input_event_name: str = Field(description="Name of the input event")
+    output_event_name: Optional[str] = Field(
+        description="Name of the output event", default=None
+    )
+
+
+class StateModification(InternalDispatchEvent):
```
masci: For consistency this should be `StateModified` or `StateChanged`

src/workflows/events.py
```diff
@@ -239,4 +241,90 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+    )
+    deleted_properties: list[str] = Field(
+        default_factory=list, description="Properties removed from the state"
+    )
+    updated_properties: list[str] = Field(
+        default_factory=list, description="Properties updated within the state"
+    )
+
+
+class QueueState(InternalDispatchEvent):
```
masci: `EventsQueueChanged`?

masci (commented):
Some minor comments around events nomenclature but the main problem I see is tracking context state changes

src/workflows/context/context.py
AstraBert: Yup, I can try and work out a solution from the state class itself :)

logan-markewich: After chatting with both of @masci @AstraBert, I think we can
1. modify/use JsonSerializer to serialize what we can (just calling to_dict() on the state)
2. any top-level keys that aren't serializable, we just cast as string (similar to how a python debugger works in your ide)
3. we expose the state before and after the step runs, and do not try to calculate differences

I think this can work ok?

The other options are exposing a basic "state changed" event (which may not be helpful) or not supporting this at all

src/workflows/events.py
```diff
@@ -239,4 +241,91 @@ async def my_step(self, ev: HumanResponseEvent) -> StopEvent:
...
+        description="State of the step ('running', 'not_running', 'in_progress', 'not_in_progress', 'exited')"
+    )
+    worker_id: str = Field(description="ID of the worker that the step is running on")
+    input_event_name: str = Field(description="Name of the input event")
+    output_event_name: Optional[str] = Field(
+        description="Name of the output event", default=None
+    )
+
+
+class StateSnapshot(InternalDispatchEvent):
```
masci: Why a separate event? I would merge it with `StepStateChanged`

PR #75, opened 2025-08-25
feat: add gather in step decorator

AstraBert:
We add a `gather` property in `StepConfig` which then, in `_step_worker`, will trigger the collection of events before the execution of the step

Example usage:

```python
class TestWf(Workflow):
    @step
    async def step_one(self, ev: StartEvent, ctx: Context[WfState]) -> SomeEvent | None:
        print("Received message " + ev.message)
        ctx.send_events([SomeEvent(data="hello"), SomeEvent(data="ciao")])
    
    @step
    async def step_process(self, ev: SomeEvent, counter: Annotated[Counter, Resource(get_counter)]) -> SecondEvent:
        counter.update()
        print("Processing message " + ev.data)
        time.sleep(1)
        return SecondEvent(greeting=ev.data)

    @step(gather=[SomeEvent])
    async def step_two(self, ev: SecondEvent, counter: Annotated[Counter, Resource(get_counter)]) -> StopEvent:
        print(ev.greeting)
        time.sleep(1)
        return StopEvent(result=f"Processed {counter.messages} messages")
```

### DIFF
--- a/src/workflows/context/context.py
+++ b/src/workflows/context/context.py
@@ -830,6 +830,9 @@ class Context(Generic[MODEL_T]):
             if type(ev) not in config.accepted_events:
                 continue
 
+            if config.gather:
+                self.collect_events(ev, config.gather)
+
             if verbose and name != "_done":
                 print(f"Running step {name}")
 
--- a/src/workflows/context/context.py
+++ b/src/workflows/context/context.py
@@ -831,7 +831,7 @@ class Context(Generic[MODEL_T]):
                 continue
 
             if config.gather:
-                self.collect_events(ev, config.gather)
+                self.gather_events(ev, config.gather)
 
             if verbose and name != "_done":
                 print(f"Running step {name}")
--- a/src/workflows/context/context.py
+++ b/src/workflows/context/context.py
@@ -521,7 +521,7 @@ class Context(Generic[MODEL_T]):
             ```
         """
         if step is None:
-            for queue in self._queues.values():
+            for name, queue in self._queues.items():
                 queue.put_nowait(message)
         else:
             if step not in self._step_configs:
@@ -830,11 +830,17 @@ class Context(Generic[MODEL_T]):
             if type(ev) not in config.accepted_events:
                 continue
 
+            gathered_evs: Optional[list[Event]] = []
+
             if config.gather:
-                self.gather_events(ev, config.gather)
+                gathered_evs = self.gather_events(ev, config.gather)
+                if not gathered_evs:
+                    continue
 
             if verbose and name != "_done":
                 print(f"Running step {name}")
+
+            events_to_process = gathered_evs if gathered_evs and config.gather else [ev]
 
             # run step
             # Initialize state manager if needed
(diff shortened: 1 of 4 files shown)


src/workflows/context/context.py
```diff
@@ -830,6 +830,9 @@ async def _step_worker(
             if type(ev) not in config.accepted_events:
                 continue
 
+            if config.gather:
+                self.gather_events(ev, config.gather)
```
logan-markewich: This makes sense, but dont we need to use the output of gather_events and only execute the step if we have all the events?

tests/test_workflow_send_gather.py
```diff
@@ -71,16 +71,13 @@ async def process_email(
         counter.update(success=sent)
         return ProcessEmailEvent(sent=sent, time=time.time())
 
-    @step
+    @step(gather=[ProcessEmailEvent])
     async def output(
         self,
         ev: ProcessEmailEvent,
```
logan-markewich, line 77: Technically this would be a list of events if using gather right?

AstraBert: Mmhh, no, it should be a list of event types, since the events that have been sent are cached with the stringified event type as key and that is how we collect them with `gather_events` :))

src/workflows/context/context.py
AstraBert: Oh yeah, absolutely!

masci: Before fixing this, can you change the test so it fails? I was also suprised tests were passing despite this line ignoring the return value

AstraBert: After experimenting quite a bit this morning, I think I can confidently say that the tests passed because `send_event` work also if you do not collect events in the end, so the presence of `gather_events` (as it was here) was basically not doing anything, it wasn't forcing the step to wait for the events before executing. With the latest implementation, it does :)

src/workflows/context/context.py
```diff
@@ -521,7 +521,7 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> WorkerEvent | GatherEve
             ```
         """
         if step is None:
-            for queue in self._queues.values():
+            for name, queue in self._queues.items():
```
masci, line 524: Why this change?

src/workflows/context/context.py
```diff
@@ -830,9 +830,18 @@ async def _step_worker(
...
+
+            if config.gather:
+                gathered_evs = self.gather_events(ev, config.gather)
+                if not gathered_evs:
+                    continue
+
             if verbose and name != "_done":
                 print(f"Running step {name}")
 
+            events_to_process = gathered_evs if gathered_evs and config.gather else [ev]
```
masci, line 843: Not sure what's going on here, `gathered_events` is obviously empty if `config.gather` is false, and when `config.gather` is true we `continue`... This could be `events_to_process = gathered_evs or [ev]` I think?

src/workflows/context/context.py
```diff
@@ -830,9 +830,18 @@ async def _step_worker(
             if type(ev) not in config.accepted_events:
                 continue
 
+            gathered_evs: Optional[list[Event]] = []
```
masci, line 833: Can we use the `|` notation instead of `Optional`?

tests/test_workflow_send_gather.py
logan-markewich: sorry, I meant the type annotation. If you are using gather in the step decorator, I would expect to annotate like `evs: List[ProcessEmailEvent],` right? In other cases, it might be `evs: List[Event1 | Event2]`

src/workflows/context/context.py
logan-markewich: Yea I think because you are only gathering a single event, it makes sense it still worked

src/workflows/context/context.py
```diff
@@ -856,88 +865,94 @@ async def _step_worker(
...
+            # Process each event
+            for current_event in events_to_process:
+                kwargs: dict[str, Any] = {}
+                if config.context_parameter:
+                    kwargs[config.context_parameter] = self
+                for resource in config.resources:
+                    kwargs[resource.name] = await resource_manager.get(
+                        resource=resource.resource
+                    )
+                kwargs[config.event_name] = current_event
```
logan-markewich, line 877: this isn't quite what I expected. Wouldn't you want to run the step with all the collected events? Here you are running the step once per event

for example
`kwargs[config.event_name] = events_to_process`

src/workflows/context/context.py
```diff
@@ -856,88 +865,94 @@ async def _step_worker(
...
-            # - check if its async or not
-            # - if not async, run it in an executor
-            if asyncio.iscoroutinefunction(step):
-                retry_start_at = time.time()
-                attempts = 0
-                while True:
-                    await self.mark_in_progress(name=name, ev=ev)
-                    await self.add_running_step(name)
+            # Process each event
+            for current_event in events_to_process:
```
masci, line 869: This looks like a big change in the runtime, can you explain the logic a bit?

AstraBert:
I think it makes sense to reply to the questions in a standalone comment to avoid duplicates:

- The logic implemented does not differ from the one we are actually running for what concerns single-event steps/steps that do not have `gather` in the decorator. 
- I was not quite sure of how to run multiple events at once within the same step, so I make sure that the step collects all events, and then it runs once per collected event
- The whole logic here is pretty much the following:

```mermaid
flowchart TD
    A(Input event)
    B[Step with gather]
    C(Tries to collect all events)
    D{All expected events are collected}
    E{Not all expected events are collected}
    F(Runs the step through all the collected events)
    G(Output Event)

    A --> B
    B --> C 
    C --> D
    C --> E
    E -->|Loops until all expected events are collected| A
    D --> F
    F --> G
```

Hope this makes things clearer @masci @logan-markewich

src/workflows/context/context.py
AstraBert: Oh, sorry, this was for debugging purposes and I just forgot to restore it

PR #80, opened 2025-09-02
docs: use pydoc-markdown to generate API reference

masci:
Instead of building the reference with mkdocs, we use `pydoc-markdown` to generate a markdown file from the docstrings in the sources. mkdocs will then treat this markdown file as any other documentation page.

Why?
- This way we can version the generated API docs
- The new documentation website will benefit from getting a markdown file ready to be rendered

Implications:
- Cross-references have a different format in `pydoc-markdown` so I had to change the docstrings
- The `workflow` section under the "API Reference" in the soon-to-be old documentation website has been removed
- Right now the generated markdown file is checked in, to be decided if it's better for the new docs website to generate it on-the-fly


logan-markewich:
@masci ooc, did we do any filtering on the exposed classes? Are there classes/functions that we don't want to expose?

masci:
We're filtering out undocumented classes/functions and private classes/functions. `_Resource` is an exception because a public member depends on it (we might want to rename it).

masci:
Discussed offline, the new developer portal will be able to support the existing mkdocs-generated api reference, closing.

PR #81, opened 2025-09-04
Add utility script to export openapi.json

adrianlyjak:
Add functionality to export openapi.json, run `uv run hatch run server:openapi`


src/workflows/server/server.py
```diff
@@ -435,3 +436,33 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
             )
+
+    def openapi_schema(self) -> dict:
+        app = self.app
+        gen = SchemaGenerator(
+            {
+                "openapi": "3.0.0",
+                "info": {
+                    "title": "Workflows API",
+                    "version": "1.0.0",
```
logan-markewich: do we have an `__version__` in the lib? Would probably be nice to get an actual version here?

adrianlyjak: good point

adrianlyjak: Used `from importlib.metadata import version`, `version("llama-index-workflows")`, which should work for python 3.8+ (min compatible is 3.9, so we should be good)

PR #83, opened 2025-09-05
feat: Add TypeScript SDK generation and publishing pipeline

zhaotai:
## Summary
- Adds automated TypeScript SDK generation from the Python server's OpenAPI schema
- Creates a unified release pipeline with proper version synchronization
- Uses `@llamaindex/workflows-client` as the npm package name

### SDK Generation
- Added `scripts/generate_sdk.py` to orchestrate SDK generation using @hey-api/openapi-ts
- Configured SDK to be generated from Starlette's OpenAPI schema
- SDK artifacts are gitignored, only configuration files are committed

### CI/CD Pipeline
- **Version Validation**: New job validates Git tags match pyproject.toml version
- **Separated Jobs**: Python and TypeScript publishing are independent jobs
- **SDK CI**: Added workflow to validate SDK generation on every PR
- **Single Source of Truth**: pyproject.toml version drives both Python and SDK versions

### Dependencies
- Added `pyyaml` to server dependencies for OpenAPI schema generation
- Configured pnpm as the package manager for Node.js operations
- Added typecheck script to package.json for validation

## Test Plan
- [x] SDK generation works locally with `uv run hatch run server:sdk-generate`
- [x] Version validation correctly fails on mismatched tags (tested with act)
- [x] Version validation passes with matching tags (tested with act)
- [x] Manual workflow dispatch uses pyproject.toml version (tested with act)
- [ ] Release workflow publishes both packages successfully
- [ ] SDK CI validates generation on PR

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -1,34 +1,240 @@
-name: Publish to PyPI and GitHub
+name: Publish Release
 
 on:
   workflow_dispatch:
+    inputs:
+      publish_python:
+        description: 'Publish Python package to PyPI'
+        required: false
+        default: true
+        type: boolean
+      publish_sdk:
+        description: 'Publish TypeScript SDK to npm'
+        required: false
+        default: true
+        type: boolean
   push:
     tags:
       - "v*"
 
 jobs:
-  build-n-publish:
-    name: Build and publish to PyPI
+  # Job 1: Validate versions
+  validate-version:
+    name: Validate Version Consistency
     if: github.repository == 'run-llama/workflows-py'
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      tag: ${{ steps.version.outputs.tag }}
+    
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Extract and validate version
+        id: version
+        run: |
+          # Extract version from pyproject.toml (single source of truth)
+          PYPROJECT_VERSION=$(grep '^version' pyproject.toml | cut -d'"' -f2)
+          echo "version=$PYPROJECT_VERSION" >> $GITHUB_OUTPUT
+          echo "PyProject version: $PYPROJECT_VERSION"
+          
+          # If this is a tag push, validate it matches
+          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
+            TAG="${GITHUB_REF#refs/tags/}"
+            echo "tag=$TAG" >> $GITHUB_OUTPUT
+            
+            # Remove 'v' prefix if present for comparison
+            TAG_VERSION="${TAG#v}"
+            
+            echo "Git tag: $TAG (version: $TAG_VERSION)"
+            
+            # Validate that tag matches pyproject.toml version
+            if [ "$PYPROJECT_VERSION" != "$TAG_VERSION" ]; then
+              echo "❌ Error: Tag version ($TAG_VERSION) does not match pyproject.toml version ($PYPROJECT_VERSION)"
+              echo "Please ensure the Git tag matches the version in pyproject.toml"
+              exit 1
+            fi
+            
+            echo "✅ Version validation passed: $PYPROJECT_VERSION"
+          else
+            echo "Manual workflow dispatch - using version from pyproject.toml"
+            echo "tag=" >> $GITHUB_OUTPUT
+          fi
+
+  # Job 2: Build and publish Python package
+  publish-python:
+    name: Publish Python Package to PyPI
+    needs: validate-version
+    if: |
+      github.repository == 'run-llama/workflows-py' &&
+      (github.event_name == 'push' || github.event.inputs.publish_python == 'true')
     runs-on: ubuntu-latest
     permissions:
       contents: write
-
+    
     steps:
       - uses: actions/checkout@v4
-
+      
       - name: Install uv
         uses: astral-sh/setup-uv@v6
-
-      - name: Build and publish
+      
+      - name: Build Python package
+        run: uv build
+      
+      - name: Publish to PyPI
         env:
           UV_PUBLISH_TOKEN: ${{ secrets.PYPI_TOKEN }}
-        run: |
-          uv build
-          uv publish
+        run: uv publish
+      
+      - name: Upload Python artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-dist
+          path: dist/
 
+  # Job 3: Build and publish TypeScript SDK
+  publish-sdk:
+    name: Publish TypeScript SDK to npm
+    needs: validate-version
+    if: |
+      github.repository == 'run-llama/workflows-py' &&
+      (github.event_name == 'push' || github.event.inputs.publish_sdk == 'true')
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write  # For npm provenance
+    
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+      
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+      
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 9
+      
+      - name: Install Python dependencies with uv
+        run: |
+          uv sync --extra server
+      
+      - name: Update SDK version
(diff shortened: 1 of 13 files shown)


.github/workflows/publish_release.yml
```diff
@@ -1,16 +1,74 @@
...
+      version: ${{ steps.version.outputs.version }}
+      tag: ${{ steps.version.outputs.tag }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract and validate version
+        id: version
+        run: |
+          # Extract version from pyproject.toml (single source of truth)
```
adrianlyjak: putting complex scripts embedded directly in github actions makes things really hard to test

.github/workflows/publish_release.yml
```diff
@@ -20,15 +78,164 @@ jobs:
...
+        run: |
+          uv sync --extra server
+
+      - name: Update SDK version
         run: |
-          uv build
-          uv publish
+          cd client-sdk
+          # Update package.json version to match Python version from validation job
+          npm pkg set version="${{ needs.validate-version.outputs.version }}"
```
adrianlyjak: actually modifying the server's API contract is going to be quite rare. This seems like a lot of thrashing of versions with no actual changes. Additionally this version won't be recorded in the code, so it'll be a little odd

.github/workflows/publish_release.yml
```diff
@@ -20,15 +78,164 @@ jobs:
...
+          body: |
+            ## 📦 Release Artifacts
+
+            ### Python Package
+            - PyPI: `pip install llama-index-workflows`
+            - Install specific version: `pip install llama-index-workflows==${{ needs.validate-version.outputs.version }}`
+
+            ### TypeScript SDK
+            - npm: `pnpm add @llamaindex/workflows-client`
+            - Install specific version: `pnpm add @llamaindex/workflows-client@${{ needs.validate-version.outputs.version }}`
```
adrianlyjak: I'm not sure we really want to advertise or publish this generated client. Its not very ergonomic, and just seems like underlying plumbing for more friendly tools

.github/workflows/sdk-ci.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Upload SDK artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: generated-sdk-${{ github.sha }}
+          path: |
+            client-sdk/src/generated/
+            client-sdk/dist/
+          retention-days: 7
+
+      - name: Version consistency check
```
adrianlyjak: this seems like a pain in the butt to have to keep this package.json in sync manually

scripts/generate_sdk.py
```diff
@@ -0,0 +1,59 @@
...
+2. Calls @hey-api/openapi-ts to generate TypeScript client
+"""
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+# Add parent directory to path to import workflows
+sys.path.insert(0, str(Path(__file__).parent.parent / "src"))
```
adrianlyjak: this doesn't look necessary. Just run it with the venv active (or with `uv run python scripts/generate_sdk.py`). You can put this in the shebang: `#!/usr/bin/env -S uv run python`

zhaotai:
> This is a fairly complex JS build process embedded in the python repo. Additionally, its kind of awkward since there's also a JS workflows repo (https://github.com/run-llama/workflows-ts), but this client certainly doesn't belong there since it doesn't have a matching server.
> 
> IMO, at least initially, this would fit better in `run-llama/llama-ui`. I'm not sure of any great use cases for the SDK beyond usage there in the react hooks. You could add an action there that pulls from this workflows-py repo, generates the latest openapi, and then triggers a generated SDK build. Then, if we want real time publishing, we can just make the workflows-py repo trigger a llama-ui action. I think there's some advantages here too with better versioning: It can check for a diff in the file, and only commit things if changed, and then just generate a changeset so that it gets bundled and bumped in the next release.

I considered `workflow-ts` repo problem. In short, it's not something we can solve in near time. If you look at our team today, basically only contractors are maintaining it. I discussed our LITS priority with Haig and Simon, all of us agree that LITS is not as important as `py` right now and even in the near future it will always be. No need to mention that when designing it, Alex was purely thinking of it from the typescript ecosystem perspective (I discussed it with Alex). So keep API contract consistency would be a heavy burden for us (Not just you and me, but the whole team) and low ROI.
That's why I put it here because it clearly indicates that this client sdk only associates with this python server. If in the future we decide to keep typescript server and python server unified. I would argue that a better solution would be place them in one repo and utilize some engineering stuffs to force/check API contract consistency (We can't rely on human especially with the team growing).

.github/workflows/publish_release.yml
zhaotai: This is to keep consistency to prevent version mismatch by accident. By unifying server package version client package version, developers are easier to recognize and maintain.

.github/workflows/sdk-ci.yml
zhaotai: why?

.github/workflows/publish_release.yml
zhaotai: I can extract them out. I feel it's pretty simple....

scripts/generate_sdk.py
zhaotai: ty.

logan-markewich:
I kind of agree with Adrian -- is this the correct home for an auto-generated SDK? Who is the main user for this?

Otherwise it seems fine, I don't have a strong opinion about where this should live at the end of the day

.github/workflows/publish_release.yml
```diff
@@ -20,15 +20,38 @@ jobs:
...
         env:
           UV_PUBLISH_TOKEN: ${{ secrets.PYPI_TOKEN }}
         run: |
           uv build
           uv publish
 
+      - name: Generate OpenAPI spec
+        run: |
+          uv sync --extra server
+          uv run python -m workflows.server.server --output openapi.json
```
adrianlyjak: you can run this with just one line as `uv run hatch run server:openapi` (there's an inline script in the pyproject.toml that's also configured with the extra requirements)

.github/workflows/publish_release.yml
```diff
@@ -20,15 +20,38 @@ jobs:
...
         uses: astral-sh/setup-uv@v6
 
+      - name: Validate version
+        if: startsWith(github.ref, 'refs/tags/')
+        run: python scripts/validate_version.py
+      
+      - name: Set version output
+        id: version
+        if: startsWith(github.ref, 'refs/tags/')
+        run: echo "version=$(grep '^version' pyproject.toml | cut -d'"' -f2)" >> $GITHUB_OUTPUT
```
adrianlyjak: alternately, should be a little more robust to syntax variation

```sh
uv run python -c "from importlib.metadata import version; print(version('llama-index-workflows'))" > $GITHUB_OUTPUT
```

scripts/validate_version.py
```diff
@@ -0,0 +1,39 @@
...
+"""Validate that release tag matches pyproject.toml version."""
+
+import os
+import sys
+from pathlib import Path
+
+
+def main() -> None:
+    # Get pyproject.toml version
+    pyproject_path = Path(__file__).parent.parent / "pyproject.toml"
```
adrianlyjak: nit: use tomlib for more valid parsing

```py
import tomllib

with open(pyproject_path) as f:
    pyproject = tomllib.load(f)
ver = tomllib['project']['version']
```

.github/workflows/sdk-ci.yml
adrianlyjak: Just in terms of updating the package.json version, seems like an unnecessary chore, given that 95%+ changes would likely be unrelated

scripts/detect_change_type.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
+"""Detect the type of change based on version tags."""
+
+import os
+import subprocess
+import sys
+from typing import Optional, Tuple
+
+
+def parse_version(version_str: str) -> Tuple[int, int, int]:
```
adrianlyjak: FYI there's some built ins for parsing and comparing versions: https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

```py
from packaging.version import Version
Version("1.0.3") > Version("0.99.88")
```

```
> v = Version("1.0.3")
>>> v.major
1
>>> v.minor
0
>>> v.micro
3
```

zhaotai: thank you! This is helpful!

PR #84, opened 2025-09-08
feat: add testing utils

AstraBert:
With this PR, we add testing utils to workflows through an async context manager (`WorkflowTestRunner`).

Example:

```python
import pytest

from workflows.events import StartEvent, StopEvent, Event, StepStateChanged, EventsQueueChanged
from workflows.testing import WorkflowTestRunner

class SecondEvent(Event):
    greeting: str

class SimpleWf(Workflow):
    @step
    async def step_one(self, ev: StartEvent, ctx: Context) -> SecondEvent | None:
            state.test = "this is a test"
        ctx.write_event_to_stream(SecondEvent(greeting="hello"))
        return SecondEvent(greeting="hello")
    @step
    async def step_two(self, ev: SecondEvent, ctx: Context) -> StopEvent:
            state.hello = "hello"
        return StopEvent(result="done")

@pytest.mark.asyncio
async def test_testing_utils():
    wf = SimpleWf()
    async with wf.run_test() as test_runner:
        assert isinstance(test_runner, WorkflowTestRunner)
        collected, freqs, res = await test_runner.run_and_collect(start_event=StartEvent(message="hi"), exclude_events=[EventsQueueChanged])
        assert len(collected) == sum([freqs[k] for k in freqs])
        assert freqs.get(SecondEvent, 0) == 1
        assert freqs.get(StopEvent, 0) == 1
        assert freqs.get(StepStateChanged, 0) == len(collected) - (freqs.get(SecondEvent, 0) + freqs.get(StopEvent, 0))
        assert str(res) == "done"
        output_event = await test_runner.send_test_event(step="step_one", event=StartEvent(message="hello"), ctx=Context(wf))
        assert isinstance(output_event, SecondEvent)
        assert output_event.greeting == "hello"
```

### DIFF
--- a/src/workflows/testing/runner.py
+++ b/src/workflows/testing/runner.py
@@ -1,0 +1,91 @@
+import inspect
+
+from typing import AsyncContextManager, Any, Callable, TYPE_CHECKING
+from types import TracebackType
+from collections import Counter
+
+if TYPE_CHECKING:
+    from workflows import Workflow
+from workflows.events import StartEvent, Event, EventType
+
+
+class WorkflowTestRunner(AsyncContextManager["WorkflowTestRunner"]):
+    def __init__(self, workflow: "Workflow"):
+        self._workflow = workflow
+
+    async def __aenter__(self) -> Any:
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> bool | None:
+        """Called when exiting the 'async with' block"""
+        return not exc_type and not exc_val and not exc_tb
+
+    async def run_and_collect(
+        self,
+        start_event: StartEvent,
+        expose_internal: bool = True,
+        exclude_events: list[EventType] | None = None,
+    ) -> tuple[list[Event], dict[EventType, int], Any]:
+        """
+        Run a workflow end-to-end and collect the events that are streamed during its execution.
+
+        Args:
+            start_event (StartEvent): The input event for the workflow
+            expose_internal (bool): Whether or not to expose internal events. Defaults to True if not set.
+            exclude_events. (list[EventType]): A list of event types to exclude from the collected events. Defaults to None if not set.
+
+        Returns:
+            A tuple containing the list of collected events, a dictionary mapping each event type with its count in the collected events and the result of the workflow run.
+
+        Example:
+            ```
+            async with wf.run_test() as test_runner:
+                collected, ev_types, result = await test_runner.run_and_collect(start_event=StartEvent(name="Adam", greeting="hello"))
+                assert len(collected) == 22
+                assert ev_types.get(StepStateChanged, 0) == 8
+                assert str(result) == "hello Adam!"
+            ```
+        """
+        handler = self._workflow.run(start_event=start_event)
+        collected_events: list[Event] = []
+        async for event in handler.stream_events(expose_internal=expose_internal):
+            if exclude_events and type(event) in exclude_events:
+                continue
+            collected_events.append(event)
+        result = await handler
+        event_freqs: dict[EventType, int] = dict(
+            Counter([type(ev) for ev in collected_events])
+        )
+        return collected_events, event_freqs, result
+
+    async def send_test_event(self, step: str, event: Event, **kwargs: Any) -> Event:
+        """
+        Send a test event to a specific step and collect the resulting output event.
+
+        Args:
+            step (str): Name of the step to send the event to
+            event (Event): Event to send
+            kwargs (Any): All necessary keyword arguments to be passed to the step (context, resources...)
+
+        Returns:
+            Output event of the step
+
+        Example:
+            ```
+            async with wf.run_test() as test_runner:
+                output_event = await test_runner.send_test_event("say_hello", SayHelloEvent(greeting="hello"), ctx=Context(wf))
+                assert isinstance(output_event, PrintHelloEvent)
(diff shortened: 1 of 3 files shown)


src/workflows/workflow.py
```diff
@@ -531,3 +532,21 @@ def _validate(self) -> bool:
             InputRequiredEvent in produced_events
             or HumanResponseEvent in consumed_events
         )
+
+    def run_test(self) -> WorkflowTestRunner:
```
masci: I would prefer the Workflow api to not depend on the test runner. This wrapper is literally one line, I think having
```py
async with WorkflowTestRunner(workflow) as test_runner:
    ...
```
is ok?

src/workflows/testing/runner.py
```diff
@@ -0,0 +1,91 @@
...
+    async def __aexit__(
+        self,
+        exc_type: Optional[type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> Optional[bool]:
+        """Called when exiting the 'async with' block"""
+        return not exc_type and not exc_val and not exc_tb
+
+    async def run_and_collect(
```
masci: I would call this just `run`, the "collection" is implicit in the fact that this is a test runner

src/workflows/testing/runner.py
```diff
@@ -0,0 +1,91 @@
...
+            collected_events.append(event)
+        result = await handler
+        event_freqs: dict[EventType, int] = dict(
+            Counter([type(ev) for ev in collected_events])
+        )
+        return collected_events, event_freqs, result
+
+    async def send_test_event(self, step: str, event: Event, **kwargs: Any) -> Event:
+        """
+        Send a test event to a specific step and collect the resulting output event.
```
masci: Why not using `ctx.send_event`? The test runner could expose the context through a property.
In case you need to exercise the step, I would just call the method directly instead, it's faster, less error prone and supports autocompletion (having to pass the name of a step does not).

masci (changes requested):
Overall looks good, left a few comments.

src/workflows/workflow.py
AstraBert: Sounds good to me :))

src/workflows/testing/runner.py
```diff
@@ -0,0 +1,64 @@
...
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> Optional[bool]:
+        """Called when exiting the 'async with' block"""
+        return not exc_type and not exc_val and not exc_tb
```
masci: Sorry for being a pester, but what's the point of having a context manager if we don't use the context? Should we run the workflow in `__aenter__` maybe? Something like:
```py
runner = WorkflowTestRunner(GreetingWorkflow())

async with runner.run(start_event=StartEvent(name="Adam", greeting="hello")) as test_run:
    assert len(test_run.collected) == 22
    assert test_run.ev_types.get(StepStateChanged, 0) == 8
    assert str(test_run.result) == "hello Adam!"
```
at that point it would make sense having a cm...

AstraBert: I tried tackling this with the latest commit. I am not super opinionated on anything used in there (I added a WorkflowTestResult to make the test result more ordered and renamed some things), and we can also simply not use this as a context manager and use it directly as a test-wrapping class.

AstraBert:
Here is the update usage pattern for the testing utils:

```python
import pytest

from workflows.events import (
    StartEvent,
    StopEvent,
    Event,
    StepStateChanged,
    EventsQueueChanged,
)
from workflows.testing import WorkflowTestRunner
from workflows.testing.runner import WorkflowTestResult


class SecondEvent(Event):
    greeting: str


class SimpleWf(Workflow):
    @step
    async def step_one(self, ev: StartEvent, ctx: Context) -> SecondEvent:
            state.test = "this is a test"
        ctx.write_event_to_stream(SecondEvent(greeting="hello"))
        return SecondEvent(greeting="hello")

    @step
    async def step_two(self, ev: SecondEvent, ctx: Context) -> StopEvent:
            state.hello = "hello"
        return StopEvent(result="done")


@pytest.mark.asyncio
async def test_testing_utils() -> None:
    wf = SimpleWf()
    runner = WorkflowTestRunner(wf)
    wf_test_run = await runner.run(
        start_event=StartEvent(message="hi"),  # type: ignore
        exclude_events=[EventsQueueChanged],
    )
    assert isinstance(wf_test_run, WorkflowTestResult)
    assert len(wf_test_run.collected) == sum(
        [wf_test_run.event_types[k] for k in wf_test_run.event_types]
    )
    assert wf_test_run.event_types.get(EventsQueueChanged, 0) == 0
    assert wf_test_run.event_types.get(SecondEvent, 0) == 1
    assert wf_test_run.event_types.get(StopEvent, 0) == 1
    assert wf_test_run.event_types.get(StepStateChanged, 0) == len(
        wf_test_run.collected
    ) - (
        wf_test_run.event_types.get(SecondEvent, 0)
        + wf_test_run.event_types.get(StopEvent, 0)
    )
    assert str(wf_test_run.result) == "done"
```

src/workflows/testing/runner.py
masci: Yeah let's remove the conrtext manager altogether at this point, I don't see the benefit honestly

AstraBert: Removed it :))

PR #87, opened 2025-09-08
Add UI to Workflow server

masci:
Opening a request for comments about how to package and ship a user interface for debugging and observability along with a Workflow Server instance. The current solution serves a single page application at the root of the Workflow Server instance.

To test it, from the root of the repo:
```sh
uv run -- python examples/server/server_example.py
```

### DIFF
--- a/src/workflows/server/static/index.html
+++ b/src/workflows/server/static/index.html
@@ -3,7 +3,7 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Workflow UI</title>
+        <title>Workflow UI</title>
     <script src="https://cdn.tailwindcss.com"></script>
 </head>
 <body class="bg-gray-100 p-4 md:p-8">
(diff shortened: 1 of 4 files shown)


src/workflows/server/static/index.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Workflow Server</title>
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/css/bootstrap.min.css" rel="stylesheet">
```
logan-markewich: small nit, but we might find it easier to use tailwind

masci: using tailwind now

logan-markewich: ❤️

logan-markewich:
Im pretty in favor of this approach

The downside I can think of is if we ever wanted to have a "hosted" version of this in the platform, it'd need a different UI codebase. But I think thats a fair tradeoff and expected

masci:
> The downside I can think of is if we ever wanted to have a "hosted" version of this in the platform, it'd need a different UI codebase. But I think thats a fair tradeoff and expected

The UI can be templatized a-la django-admin if needed, the approach would stay mostly the same, just we would need an intermediate "template rendering" step using default assets out of the box.

PR #90, opened 2025-09-10
Proposals for UI improvements

AstraBert:
Adding my proposal for UI improvement here (can also commit directly into the massi/ui branch, but did not want to overstep)

Updates (as of 9/10/25):
 
- [X] Represent events data as collapsible
- [X] Provide Input Event Schema instead of JSON textbox
- [X] Provide Output Event Schema

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -64,8 +64,8 @@ class WorkflowServer:
                 methods=["POST"],
             ),
             Route(
-                "/workflows/{name}/start-event",
-                self._get_start_event_schema,
+                "/workflows/{name}/schema",
+                self._get_events_schema,
                 methods=["GET"],
             ),
             Route(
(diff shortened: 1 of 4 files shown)


src/workflows/server/server.py
```diff
@@ -63,6 +63,11 @@ def __init__(self, middleware: list[Middleware] | None = None):
                 self._run_workflow_nowait,
                 methods=["POST"],
             ),
+            Route(
+                "/workflows/{name}/start-event",
```
masci: I wonder if we should do `"/workflows/{name}/schema"` and return both the start event and the stop event?

AstraBert: Makes sense! I think I could try drafting it also on the UI side :))

AstraBert: Now we have it!

PR #92, opened 2025-09-11
simple workflow server persistence

adrianlyjak:
Adds an optional workflow storage abstractions, such that workflows can be resumed on startup

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -775,11 +775,7 @@ class WorkflowServer:
                     isinstance(event, StepStateChanged)
                     and event.step_state == StepState.NOT_RUNNING
                 ):
-                    state = (
-                        event.context_state or handler.ctx.to_dict()
-                        if handler.ctx
-                        else None
-                    )
+                    state = handler.ctx.to_dict() if handler.ctx else None
                     if state is None:
                         logger.warning(
                             f"Context state is None for handler {handler_id}. This is not expected."


src/workflows/server/server.py
```diff
@@ -597,19 +639,104 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
+@dataclass
+class _WorkflowHandler:
+    """A wrapper around a WorkflowHandler, since the main handler's queue is being read from internally"""
+
+    handler: WorkflowHandler
+    queue: asyncio.Queue[Event]
+    task: asyncio.Task[None]
+
+    async def iter_events(self) -> AsyncGenerator[Event, None]:
+        while not self.queue.empty() or not self.task.done():
```
adrianlyjak, line 831: I think ideally, something could be moved into the workflow handler such that multiple consumers can co-exist. It would make this separate queue unnecessary

src/workflows/server/server.py
```diff
@@ -597,19 +639,104 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
-            {
-                "openapi": "3.0.0",
-                "info": {
-                    "title": "Workflows API",
-                    "version": version("llama-index-workflows"),
-                },
-            }
+    @asynccontextmanager
+    async def _lifespan(self, _: Starlette) -> AsyncGenerator[None, None]:
+        await self._initialize_active_handlers()
```
logan-markewich, line 726: So on workflow startup, this is checking the store for any incomplete runs and restarting them

src/workflows/server/server.py
```diff
@@ -597,19 +639,104 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
+                            workflow_name=workflow_name,
+                            completed=False,
+                            ctx=event.context_state,
+                        )
+                    )
+                if not isinstance(event, InternalDispatchEvent):
+                    queue.put_nowait(event)
+
+        task = asyncio.create_task(_stream_events())
+        self._handlers[handler_id] = _WorkflowHandler(handler, queue, task)
```
logan-markewich, line 815: This is storing a small helper per handler to save the events until they are requested

src/workflows/server/server.py
```diff
@@ -597,19 +639,104 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
+    def _run_workflow_handler(
+        self, handler_id: str, workflow_name: str, handler: WorkflowHandler
+    ) -> None:
+        queue: asyncio.Queue[Event] = asyncio.Queue()
+
+        async def _stream_events() -> None:
+            async for event in handler.stream_events(expose_internal=True):
+                if (
+                    isinstance(event, StepStateChanged)
+                    and event.step_state == StepState.NOT_IN_PROGRESS
```
logan-markewich: We need a specific internal event to save the state on (it may or may not be this one, not totally sure fyi)

src/workflows/server/server.py
adrianlyjak: these are great, I'm going to add to the code as comments, lol

src/workflows/server/server.py
```diff
@@ -597,19 +639,104 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
+            )
 
-        return gen.get_schema(app.routes)
+    def _run_workflow_handler(
+        self, handler_id: str, workflow_name: str, handler: WorkflowHandler
+    ) -> None:
+        queue: asyncio.Queue[Event] = asyncio.Queue()
+
+        async def _stream_events() -> None:
+            async for event in handler.stream_events(expose_internal=True):
```
logan-markewich, line 773: we probably need an error state for runs (both when persisting them, and we also don't want to restart error'd out runs)

src/workflows/server/server.py
adrianlyjak: Yes, exactly. Looking deeper, this is probably not the right one. I was going off of the events that had a not None value for the context dict, but that is maybe irrelevant. NOT_RUNNING looks like maybe the better one, and I can just grab the latest context from the handler.

src/workflows/server/abstract_workflow_store.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Optional, List, Any
+from dataclasses import dataclass
+from pydantic import BaseModel
+
+
+@dataclass()
+class HandlerQuery:
+    handler_id: Optional[List[str]] = None
+    workflow_name: Optional[List[str]] = None
+    completed: Optional[bool] = None
```
zhaotai: Can we use `status` instead? So that we can extend easily in the future.
Now we have 3 status: running, completed, failed. Maybe in the future we will have a `queued` or `dormant` since we will have `resume`

src/workflows/server/abstract_workflow_store.py
```diff
@@ -0,0 +1,35 @@
...
+from abc import abstractmethod, ABC
+from typing import Optional, List, Any
+from dataclasses import dataclass
+from pydantic import BaseModel
+
+
+@dataclass()
+class HandlerQuery:
+    handler_id: Optional[List[str]] = None
+    workflow_name: Optional[List[str]] = None
```
zhaotai: names?

src/workflows/server/server.py
zhaotai: +1

src/workflows/server/server.py
adrianlyjak: agreed, terry has the same point. (I think it'll not re-run them, just show them as completed)

src/workflows/server/server.py
adrianlyjak: looking at the internals closer, I think it might make more sense to just handle the historical log and multiple stream_result consumers here, rather than bloating the internals and messing with the serialized context

src/workflows/server/abstract_workflow_store.py
adrianlyjak: I had it singular because its more about matching the `workflow_name` field to the items within the list. Updated instead to reflect the "IN" semantics: `workflow_name_in` 🤷

src/workflows/server/abstract_workflow_store.py
adrianlyjak: updated

pyproject.toml
```diff
@@ -46,3 +47,6 @@ dependencies = ["pyyaml>=6.0.2"]
 
 [tool.hatch.envs.server.scripts]
 openapi = "python -m workflows.server.server --output openapi.json"
+
+[tool.pytest.ini_options]
+asyncio_mode = "auto"
```
adrianlyjak, line 52: async generator was not working as a fixture without this

pyproject.toml
```diff
@@ -11,7 +11,8 @@ dev = [
   "httpx>=0.25.0",
   "hatch>=1.14.1",
   "pyyaml>=6.0.2",
-  "packaging>=21.0"
+  "packaging>=21.0",
+  "pytest-xdist>=3.8.0"
```
adrianlyjak, line 15: Not necessary, but wanted to run tests faster, there's a few that dawdle, this 4x speeds up the test time for me

src/workflows/server/server.py
```diff
@@ -460,10 +505,10 @@ async def _stream_events(self, request: Request) -> StreamingResponse:
         sse = request.query_params.get("sse", "false").lower() == "true"
         media_type = "text/event-stream" if sse else "application/x-ndjson"
 
-        async def event_stream(handler: WorkflowHandler) -> AsyncGenerator[str, None]:
+        async def event_stream(handler: _WorkflowHandler) -> AsyncGenerator[str, None]:
             serializer = JsonSerializer()
 
-            async for event in handler.stream_events(expose_internal=True):
```
adrianlyjak, line 466: Looks like with the UI change we're now always exposing internal events? Was that intentional? This seems like perhaps something that should be a query parameter

PR #94, opened 2025-09-12
feat: support more start event input formats in server API

adrianlyjak:
the client start event format is a bit awkward to work with, and doesn't necessarily need to be this way.
```json
{"start_event":"{\"__is_pydantic\":true,\"value\":{\"file_id\":\"ec6a93ce-1cc9-4610-b684-d706a6212ac0\"},\"qualified_name\":\"test_proj.process_file.FileEvent\"}"}
```

1. we can support passing in an object directly, rather than a string, 
2. the workflow has knowledge of the start event, so we can just use that

This change adds support for plain objects in addition to the previous format:

```json
{"start_event": {"file_id":"ec6a93ce-1cc9-4610-b684-d706a6212ac0"}}
```

It also supports objects with the __is_pydantic fields.

Additionally, this adds validation that the start event is a conformant start event. Otherwise you'd need to call the result before getting this error message. (I think there's some further bugs where if you start streaming a workflow after it has failed, it just hangs indefinitely)

### DIFF
--- a/tests/server/test_server_endpoints.py
+++ b/tests/server/test_server_endpoints.py
@@ -44,13 +44,6 @@ async def async_client(
                 pass
 
 
-# Add a test for:
-# - passing in a start event as a string, without the __is_pydantic and qualified_name fields
-# - passing in a start event as a dict, with the __is_pydantic and qualified_name fields
-# - passing in a start event as a dict, without the __is_pydantic and qualified_name fields
-# - passing in an event type that does not conform to the workflow's start event class
-
-
 @pytest.mark.asyncio
 async def test_run_workflow_with_start_event_str_plain(
     async_client: AsyncClient,


src/workflows/context/serializers.py
```diff
@@ -60,8 +60,18 @@ class JsonSerializer(BaseSerializer):
         - [PickleSerializer][workflows.context.serializers.PickleSerializer]
     """
 
-    def _serialize_value(self, value: Any) -> Any:
-        """Helper to serialize a single value."""
+    def serialize_value(self, value: Any) -> Any:
```
adrianlyjak, line 63: these seem safe to expose, as long as the caller is aware of the difference, similar to pydantic's `model_validate` vs `model_validate_json`

tests/server/test_server_endpoints.py
```diff
@@ -44,6 +44,80 @@ async def async_client(
                 pass
 
 
+# Add a test for:
+# - passing in a start event as a string, without the __is_pydantic and qualified_name fields
+# - passing in a start event as a dict, with the __is_pydantic and qualified_name fields
+# - passing in a start event as a dict, without the __is_pydantic and qualified_name fields
+# - passing in an event type that does not conform to the workflow's start event class
```
logan-markewich: nit: remove comments

adrianlyjak: derp

src/workflows/server/server.py
```diff
@@ -645,19 +645,34 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
-                    start_event = serializer.deserialize(start_event_str)
+                    start_event = (
+                        serializer.deserialize(start_event_data)
+                        if isinstance(start_event_data, str)
+                        else serializer.deserialize_value(start_event_data)
+                    )
+                    if isinstance(start_event, dict):
+                        start_event = workflow.start_event_class.model_validate(
+                            start_event
+                        )
```
masci, line 663: Can we assume `isinstance(start_event, dict)` is true if `start_event` comes from `deserialize_value`? If that's true, I would merge these two conditionals and spare one call to `isinstance()`, e.g.
```py
if isinstance(start_event_data, str):
    start_event = serializer.deserialize(start_event_data)
else:
    serializer.deserialize_value(start_event_data)
    start_event = workflow.start_event_class.model_validate ...
```

adrianlyjak: I think that will work, given that this is true:

```python
>>> class MyModel(pydantic.BaseModel):
...     foo: int
...
>>> MyModel.model_validate(MyModel(foo=3))
MyModel(foo=3)
```

adrianlyjak: (trying to support the full matrix of possibilities, since if its being flexible, a little weird if one doesn't work)

|      | string | dict |
|--- |-------|-----|
|is_pydantic| x | x |
| bare object | x | x |

adrianlyjak: wait, no, that breaks the bare string, which is what the llama-ui hook is currently sending (the old llama deploy format)

|      | string | dict |
|----|-------|-----|
| is_pydantic |  ✅   |  ✅  |
| bare object |  💀   | ✅ |
```

PR #101, opened 2025-09-16
Add better error handling and lifecycle handling to workflow server

adrianlyjak:
- Adds retries to workflow server persistence, and if it fully fails, cancel/fail the workflow
- Adds workflow error exception logging
- Shims in a task reference such that the persistence tests can wait for things to quiet down before quitting
- Makes WorkflowServer into a context manager. This is related to alleviating some technical debt in LlamaDeploy before that can be paid down, but it also makes it simpler and more obvious how to tear down in tests. Context manager isn't stricly required, but I can also manage this by making a start/stop interface public

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -9,6 +9,7 @@ import json
 import logging
 from importlib.metadata import version
 from pathlib import Path
+from types import TracebackType
 from typing import Any, AsyncGenerator
 
 import uvicorn
@@ -38,6 +39,7 @@ from workflows.server.abstract_workflow_store import (
     EmptyWorkflowStore,
     HandlerQuery,
     PersistentHandler,
+    Status,
 )
 from .utils import nanoid
 
@@ -47,8 +49,11 @@ logger = logging.getLogger()
 class WorkflowServer:
     def __init__(
         self,
+        *,
         middleware: list[Middleware] | None = None,
         workflow_store: AbstractWorkflowStore = EmptyWorkflowStore(),
+        # retry/backoff seconds for persisting the handler state in the store after failures. Configurable mainly for testing.
+        persistence_backoff: list[float] = [0.5, 3],
     ):
         self._workflows: dict[str, Workflow] = {}
         self._contexts: dict[str, Context] = {}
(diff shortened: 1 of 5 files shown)


logan-markewich (approved):
I think the context manager syntax is a little unconventional for a server? But otherwise this looks good 

(I think I understand why we went this way, makes it easier when wrapping into other servers I think?)

src/workflows/server/server.py
```diff
@@ -784,7 +794,38 @@ def _run_workflow_handler(
...
         queue: asyncio.Queue[Event] = asyncio.Queue()
 
-        async def _stream_events() -> None:
+        async def _stream_events(handler: WorkflowHandler) -> None:
+            async def checkpoint(status: Status) -> None:
+                if not handler.ctx:
+                    return
+                ctx = handler.ctx.to_dict()
+                backoffs = list(self._persistence_backoff)
+                while True:
```
logan-markewich, line 811: Is tenacity a better choice here?

adrianlyjak: Didn’t want to add a new dependency

adrianlyjak:
> I think the context manager syntax is a little unconventional for a server? But otherwise this looks good
> 
> (I think I understand why we went this way, makes it easier when wrapping into other servers I think?)

Yeah, I have some doubts about whether this is actually a good interface. Kind of wondering if it should more just expose a context manager method (like the _lifecycle) rather than being the context manager

src/workflows/workflow.py
```diff
@@ -420,13 +415,17 @@ async def _run_workflow() -> None:
 
                 result.set_result(ctx._retval)
             except Exception as e:
-                result.set_exception(e)
+                if not result.done():
+                    result.set_exception(e)
```
logan-markewich, line 419: Why this check? If theres an exception shouldn't it be set regardless?

src/workflows/server/server.py
logan-markewich: Fair

src/workflows/workflow.py
adrianlyjak: This will crash instead with an invalid state error if it’s already done

adrianlyjak: (Showed up in the test shutdown logs)

logan-markewich: Thanks for clarifying!

PR #104, opened 2025-09-18
feat: workflow visualization

AstraBert:
Tackling milestones 1 and 2 on the workflow UI roadmap (workflow visualization)

### DIFF
--- a/src/workflows/server/representation_utils.py
+++ b/src/workflows/server/representation_utils.py
@@ -1,5 +1,5 @@
-from dataclasses import dataclass
-from typing import List, Optional
+from dataclasses import dataclass, asdict
+from typing import List, Optional, Any
 
 from workflows.events import (
     StopEvent,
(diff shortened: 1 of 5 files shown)


src/workflows/server/static/main.js
```diff
@@ -131,9 +135,23 @@ document.addEventListener('DOMContentLoaded', () => {
                     if (line.trim() === '') continue;
                     try {
                         const eventData = JSON.parse(line);
+                        highlightNode(eventData.qualified_name.replace("__main__.", "").replace("workflows.events.", ""))
```
logan-markewich: This seems a little brittle (I dont think its guaranteed to be main, and it might also have more paths depending on where in the source code that the event is defined)

src/workflows/server/representation_utils.py
```diff
@@ -0,0 +1,246 @@
...
+            else:
+                edges.append(DrawWorkflowEdge(event_type.__name__, step_name))
+
+            if issubclass(event_type, HumanResponseEvent):
+                edges.append(DrawWorkflowEdge("external_step", event_type.__name__))
+
+    return DrawWorkflowGraph(nodes=nodes, edges=edges)
+
+
+def workflow_to_cytoscape(workflow_graph: DrawWorkflowGraph) -> dict:
```
logan-markewich: Since this is going into WorkflowServer, it might be better to keep this a generic JSON representation, and keep the cytoscape specific code in yhe frontend. This way, other frontends can use this without using or knowing about cytoscape

AstraBert: Backend server now returns a generic JSON and the conversion to a CytoScape-compatible object is done by the JS script :))

src/workflows/server/static/main.js
AstraBert: Should be resolved now :))

PR #107, opened 2025-09-22
feat: switch to SSE by default

adrianlyjak:
I'm looking at integrating SSE support to llama-ui and the generated workflows-client. For hey-api to generate an SSE stream client, it requires the response format to be explicitly just SSE. Seems reasonable to have this as the default, as its more of a standard. When using SSE, browser are able to render the response stream as data comes in

firefox example:

image

Changes:
- Changes the SSE response format. Using the event name as a `type:` field is pretty restrictive to clients. To subscribe to "non default" sse events (those that have a type). You need to add an event handler for each event type. an `eventSource.addEventListener("<qualified event name>", (event) => {})`. This renders the SSE events largely unusable. Instead, events are emitted very similar to the ndjson, with the fully serialized event
- Updates the debugger ui to use sse events
- updates the sse query param default from false to true. This is a breaking change, but I don't think anyone is yet using this server, and our internal clients are all providing the sse true/false param


examples/server/server_example.py
```diff
@@ -79,7 +79,7 @@ class ProcessingWorkflow(Workflow):
 
     @step
     async def process(self, ctx: Context, ev: StartEvent) -> StopEvent:
-        items = getattr(ev, "items", ["item1", "item2", "item3", "item4", "item5"])
+        items = getattr(ev, "things", ["item1", "item2", "item3", "item4", "item5"])
```
adrianlyjak, line 82: "items" is a method on StartEvent since it's dict-like, and this blows up trying to iterate over a method

src/workflows/server/static/main.js
```diff
@@ -120,74 +121,94 @@ document.addEventListener('DOMContentLoaded', () => {
     async function streamEvents(handlerId) {
         eventStreams[handlerId] = [];
         try {
-            const response = await fetch(`/events/${handlerId}`);
```
adrianlyjak, line 123: this change is all very vibe-coded, but I think that's the norm in this file. In general, I'm concerned about adding more to this without it falling down under its own weight--I would love it to integrate with @llamaindex/ui, it has a lot of useful patterns, and we could easily add a full ui "app" to that monorepo while building out useful workflow centric components

logan-markewich: @adrianlyjak but we can have our [very own monofile](https://llama-index.slack.com/archives/C01A4LY0FT3/p1756484109870479) ❤️ 

For real though, we can always split this into multiple files if needed. I'm not 100% sure what it would mean to integrate with llamaindex/ui, but open to suggestions!

src/workflows/server/server.py
```diff
@@ -585,30 +585,44 @@ async def _stream_events(self, request: Request) -> StreamingResponse:
...
             content:
-              application/x-ndjson:
-                schema:
-                  type: string
-                  description: Newline-delimited JSON stream of events.
+              # Intentionally omits the application/x-ndjson content type to simplify client generation
+              # application/x-ndjson:
+              #   schema:
+              #     type: string
+              #     description: Newline-delimited JSON stream of events.
```
logan-markewich: should we just remove this string?

src/workflows/server/static/main.js
logan-markewich, line 1: I'm going to guess you quickly popped open the UI to make sure it still works?

logan-markewich (approved):
I'm in favor

src/workflows/server/static/main.js
logan-markewich: I guess we could make `llamaindex/ui` a library of useful functions for interacting with the WorkflowServer, and build both useWorkflow and this app on top of those functions?

adrianlyjak: Yeah, it already exists! These hooks are all compatible with the workflow server (its the reason I'm making these changes). 
docs: https://developers.llamaindex.ai/python/cloud/llamaagents/ui-hooks/
package: https://www.npmjs.com/package/@llamaindex/ui

src/workflows/server/static/main.js
adrianlyjak: yup!

src/workflows/server/server.py
adrianlyjak: Do you have a preference? I deleted it, and felt bad because it was lying. Did this as a compromise

logan-markewich: Hmm, since its commented out, I'm not sure what impact this has on the generated API reference?

src/workflows/server/static/main.js
logan-markewich: I would probably wait until we have the UI fully working before doing this (since the requirements are very much in flux before that)

src/workflows/server/server.py
adrianlyjak: No impact. Just removed it now. Au Revior 👋

PR #109, opened 2025-09-23
Add python workflow client

AstraBert:
Attempting the creation of a workflow client

### DIFF
--- a/src/workflows/client/client.py
+++ b/src/workflows/client/client.py
@@ -1,7 +1,14 @@
 import httpx
 import json
 
-from typing import Literal, Any, Union, AsyncGenerator, AsyncIterator, Optional
+from typing import (
+    Any,
+    Union,
+    AsyncGenerator,
+    AsyncIterator,
+    Optional,
+    overload,
+)
 from contextlib import asynccontextmanager
 from workflows.context.serializers import JsonSerializer
 from workflows.events import StartEvent, Event
@@ -16,30 +23,35 @@ from workflows.protocol import (
 
 
 class WorkflowClient:
+    @overload
+    def __init__(self, *, httpx_client: httpx.AsyncClient): ...
+    @overload
     def __init__(
         self,
-        protocol: Optional[Literal["http", "https"]] = None,
-        host: Optional[str] = None,
-        port: Optional[int] = None,
-        timeout: Optional[int] = None,
-        httpx_kwargs: Optional[dict[str, Any]] = None,
+        *,
+        base_url: str,
+    ): ...
+
+    def __init__(
+        self,
+        *,
+        httpx_client: Union[httpx.AsyncClient, None] = None,
+        base_url: Union[str, None] = None,
     ):
-        # TODO: middleware-related logic
-        self.protocol = protocol or "http"
-        self.host = host or "localhost"
-        self.port = port or 8000
-        self.timeout = timeout or 600
-        self.httpx_kwargs = httpx_kwargs or {}
-        # TODO: add some basic TLS/verification and auth features
+        if httpx_client is None and base_url is None:
+            raise ValueError("Either httpx_client or base_url must be provided")
+        if httpx_client is not None and base_url is not None:
+            raise ValueError("Only one of httpx_client or base_url must be provided")
+        self.httpx_client = httpx_client
+        self.base_url = base_url
 
     @asynccontextmanager
     async def _get_client(self) -> AsyncIterator[httpx.AsyncClient]:
-        async with httpx.AsyncClient(
-            base_url=self.protocol + "://" + self.host + ":" + str(self.port),
-            timeout=self.timeout,
-            **self.httpx_kwargs,
-        ) as client:
-            yield client
+        if self.httpx_client:
+            yield self.httpx_client
+        else:
+            async with httpx.AsyncClient(base_url=self.base_url or "") as client:
+                yield client
 
     async def is_healthy(self) -> HealthResponse:
         """
--- a/examples/client/base/workflow_client.py
+++ b/examples/client/base/workflow_client.py
@@ -15,7 +15,7 @@ class InputNumbers(StartEvent):
 
 
 async def main() -> None:
-    client = WorkflowClient(protocol="http", host="localhost", port=8000)
+    client = WorkflowClient(base_url="http://localhost:8000")
     workflows = await client.list_workflows()
     print("===== AVAILABLE WORKFLOWS ====")
     print(workflows)
--- a/examples/client/human_in_the_loop/workflow_client_hitl.py
+++ b/examples/client/human_in_the_loop/workflow_client_hitl.py
@@ -16,7 +16,7 @@ class OutEvent(StopEvent):
 
 
 async def main() -> None:
-    client = WorkflowClient(protocol="http", port=8000, host="localhost")
+    client = WorkflowClient(base_url="http://localhost:8000")
     handler = await client.run_workflow_nowait("human")
     handler_id = handler.handler_id
     print(handler_id)
--- a/src/workflows/protocol/__init__.py
+++ b/src/workflows/protocol/__init__.py
@@ -52,7 +52,7 @@ class WorkflowEventsListResponse(BaseModel):
 
 
 class WorkflowGraphResponse(BaseModel):
-    graph: "WorkflowGraphNodeEdges"
+    graph: WorkflowGraphNodeEdges
 
 
 class WorkflowGraphNode(BaseModel):
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -1,6 +1,6 @@
 import pytest
 
-from httpx import ASGITransport
+from httpx import ASGITransport, AsyncClient
 from workflows.server.server import WorkflowServer
 from workflows.client import WorkflowClient
 from .greeting_workflow import greeting_wf, InputEvent, OutputEvent
@@ -16,7 +16,8 @@ def server() -> WorkflowServer:
 @pytest.fixture()
 def client(server: WorkflowServer) -> WorkflowClient:
     transport = ASGITransport(server.app)
-    return WorkflowClient(httpx_kwargs={"transport": transport})
+    httpx_client = AsyncClient(transport=transport, base_url="http://test")
+    return WorkflowClient(httpx_client=httpx_client)
 
 
 @pytest.mark.asyncio


pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
 
 [project.optional-dependencies]
 server = ["starlette>=0.39.0", "uvicorn>=0.32.0"]
+client = ["httpx>=0.28.1"]
```
adrianlyjak: we should bound this to `<1`. Looks like they're working on some breaking changes

src/workflows/client/client.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        Run the workflow in the background.
+
+        Args:
+            start_event (Union[StartEvent, dict[str, Any], None]): start event class or dictionary representation (optional, defaults to None and get passed as an empty dictionary if not provided).
+            context: Context or serialized representation of it (optional, defaults to None if not provided)
+            **kwargs: Any number of keyword arguments that would be passed on as additional keyword arguments to the workflow.
+
+        Returns:
+            str: ID of the handler running the workflow
```
adrianlyjak: Maybe return a typed dict for the workflow handler dict that @zhaotai is adding? https://github.com/run-llama/workflows-py/pull/110

src/workflows/client/client.py
```diff
@@ -0,0 +1,342 @@
...
+        async with self._get_client() as client:
+            response = await client.post(
+                f"/workflows/{workflow_name}/run-nowait", json=request_body
+            )
+
+            response.raise_for_status()
+
+            return response.json()["handler_id"]
+
+    async def _stream_events_sse(
```
adrianlyjak: I don't see a reason to support both of these. I think for python, the ndjson is a lot simpler

src/workflows/client/client.py
```diff
@@ -0,0 +1,342 @@
...
+                        # Handler completed, no more events
+                        return
+
+                    response.raise_for_status()
+
+                    async for line in response.aiter_lines():
+                        if line.strip():  # Skip empty lines
+                            try:
+                                event = json.loads(line.replace("\n", ""))
+                                yield event.get("value", {})
```
adrianlyjak: this shouldn't just emit the `value`. Ignoring the `qualified_name` is throwing out valueable information, otherwise the caller doesn't know exactly what type this is supposed to be. Especially with workflows, where there's lots of meta programming based on classes (e.g. there's often events without even any values). 

We could also perhaps have an option to plumb it through the JsonSerializer in case they happen to have the same python class around? Not sure what that would really look like though.

src/workflows/client/client.py
```diff
@@ -0,0 +1,342 @@
...
+                raise TimeoutError(
+                    f"Timeout waiting for events from handler {handler_id}"
+                )
+            except httpx.RequestError as e:
+                raise ConnectionError(f"Failed to connect to event stream: {e}")
+
+    async def stream_events(
+        self,
+        handler_id: str,
+        event_callback: Callable[[dict[str, Any]], Any] | None = None,
```
adrianlyjak: why isn't this just an async generator too? That seems easier than having a callback

src/workflows/client/client.py
```diff
@@ -0,0 +1,342 @@
...
+            handler_id (str): ID of the handler of the running workflow to send the event to
+            event (Event | dict[str, Any] | str): Event to send, represented as an Event object, a dictionary or a serialized string.
+            step (str | None): Step to send the event to (optional, defaults to None)
+
+        Returns:
+            bool: Success status of the send operation
+        """
+        if isinstance(event, Event):
+            try:
+                event = event.model_dump_json()
```
adrianlyjak: this won't work. We need to use that `JsonSerializer`, otherwise the server will just get this back of json values, and not know what it's supposed to deserialize to. I think rather than requiring a `dict[str, Any]`, we should have something simple like a TypedDict or yet another pydantic model to represent the serialized state. (the thing with the `qualified_name`)

src/workflows/client/client.py
```diff
@@ -0,0 +1,342 @@
...
+            Any: Result of the workflow
+        """
+        async with self._get_client() as client:
+            response = await client.get(f"/results/{handler_id}")
+            response.raise_for_status()
+
+            if response.status_code == 202:
+                return
+
+            return response.json()["result"]
```
adrianlyjak: @zhaotai is also updating this API to return the full handler, so maybe don't throw out the rest of that info

adrianlyjak:
IMO, I think we'll need a low level basic API client like this either way, so this is great, but I kind of think we should treat it as private, and rather expose an "easy to use" interface that looks more like the handler response from a `workflow.run`. That also maybe frees us up to make more internal changes as long as we keep the outer interface more stable

adrianlyjak:
Also worth considering potentially generating a client for the lower API integration (with something like https://github.com/openapi-generators/openapi-python-client). The main advantage there is probably keeping us honest as types change while we develop the API.

We have the OpenAPI spec easily exportable already.

src/workflows/client/client.py
```diff
@@ -0,0 +1,288 @@
...
+            raise ConnectionError(f"Failed to connect to event stream: {e}")
+
+    async def get_workflow_handlers(self) -> list[Handler]:
+        response = await get_handlers(client=self._client)
+        if response:
+            return response.handlers
+        else:
+            raise ValueError("Response was not properly generated")
+
+    async def get_workflow_result(self, handler: Handler) -> Any:
```
adrianlyjak: shouldn't this just be a handler id? as input, and return a Handler as a result?

src/workflows/client/client.py
```diff
@@ -0,0 +1,288 @@
...
+            raise ValueError("Response was not properly generated")
+
+    async def get_workflow_events(
+        self,
+        handler: Handler,
+    ) -> AsyncGenerator[dict[str, Any], None]:
+        """
+        Stream events using newline-delimited JSON format
+        """
+        url = f"/events/{handler.handler_id}?sse=false"
```
adrianlyjak: nit: url encode the handler in case its some garbage string that has slashes/question marks in it

src/workflows/client/client.py
```diff
@@ -0,0 +1,288 @@
...
+            ),
+        )
+        if isinstance(response, Handler):
+            return response
+        else:
+            raise ValueError("Response was not properly generated")
+
+    async def get_workflow_events(
+        self,
+        handler: Handler,
```
adrianlyjak: shouldn't this just be a string id?

adrianlyjak (commented):
added some small questions around handler parameters. @AstraBert how do you feel about the generated client? I know I suggested it, but wondering if its helping or hurting (its astonishing how much code it generates!)

src/workflows/client/client.py
AstraBert: Yeah, it should, but I thought it would be easier and more future-proof passing the handler (maybe in the future we want to use more parameters from it). But it can be easily reverted to handler id, I also think it may be a little bit bloaty to put the entire handler there

src/workflows/client/client.py
AstraBert: I did not know it, I'll sanitize it later today along with reverting to pass onlu the handler id

AstraBert:
@adrianlyjak personally not a super fan of this last iteration, I feel like we could do more with less code (previous iteration :))

AstraBert:
Reverted back to the initial implementation with some adjustments, @adrianlyjak :)

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+        Returns:
+            bool: True if the workflow server is healthy, false if not
+        """
+        async with self._get_client() as client:
+            response = await client.get("/health")
+        if response.status_code == 200:
+            return response.json().get("status", "") == "healthy"
+        return False
+
+    async def ping(self) -> float:
```
adrianlyjak: nit: not sure I see value in having this predefined in the client

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Args:
+            start_event (Union[StartEvent, dict[str, Any], None]): start event class or dictionary representation (optional, defaults to None and get passed as an empty dictionary if not provided).
+            context: Context or serialized representation of it (optional, defaults to None if not provided)
+            **kwargs: Any number of keyword arguments that would be passed on as additional keyword arguments to the workflow.
+
+        Returns:
+            dict[str, Any]: JSON representation of the handler running the workflow
+        """
+        if isinstance(start_event, StartEvent):
```
adrianlyjak: nit: nice to have some shared functions to consolidate this duplication

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+            except Exception as e:
+                raise ValueError(
+                    f"Impossible to serialize the start event because of: {e}"
+                )
+        if isinstance(context, Context):
+            try:
+                context = context.to_dict()
+            except Exception as e:
+                raise ValueError(f"Impossible to serialize the context because of: {e}")
+        request_body = {
```
adrianlyjak: nit: I think we should build out some shared `TypedDict` models between client/server to lightly type things and make type discrepancies obvious

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+            response.raise_for_status()
+
+            return response.json()["result"]
+
+    async def run_workflow_nowait(
+        self,
+        workflow_name: str,
+        start_event: Union[StartEvent, dict[str, Any], None] = None,
+        context: Union[Context, dict[str, Any], None] = None,
+        **kwargs: Any,
```
adrianlyjak: I've never actually found a use case for these kwargs. I'd recommend hiding them from the interface to avoid breaking changes, and if we do need to introduce them, use `Unpack[TypedDict]` types so that the exact set of supported options is obvious and documented

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+
+            response.raise_for_status()
+
+            return response.json()["result"]
+
+    async def run_workflow_nowait(
+        self,
+        workflow_name: str,
+        start_event: Union[StartEvent, dict[str, Any], None] = None,
+        context: Union[Context, dict[str, Any], None] = None,
```
adrianlyjak, line 132: this and `run` are missing the new `handler_id` option

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+                f"/workflows/{workflow_name}/run-nowait", json=request_body
+            )
+
+            response.raise_for_status()
+
+            return response.json()
+
+    async def get_workflow_events(
+        self,
+        handler_id: str,
```
adrianlyjak, line 174: needs the new include_internal option

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+            handler_id (str): ID of the handler of the running workflow to send the event to
+            event (Event | dict[str, Any] | str): Event to send, represented as an Event object, a dictionary or a serialized string.
+            step (str | None): Step to send the event to (optional, defaults to None)
+
+        Returns:
+            bool: Success status of the send operation
+        """
+        if isinstance(event, Event):
+            try:
+                event = event.model_dump_json()
```
adrianlyjak: this is the wrong way to serialize. These need to be the `JsonSerializer` serialized values, otherwise the server will not know what type to deserialize to

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+            Any: Result of the workflow
+        """
+        async with self._get_client() as client:
+            response = await client.get(f"/results/{handler_id}")
+            response.raise_for_status()
+
+            if response.status_code == 202:
+                return
+
+            return response.json()["result"]
```
adrianlyjak: I think we should expose the full response TypedDict (already defined for the server), and return that. That way you can get the current status, etc.

src/workflows/client/client.py
```diff
@@ -0,0 +1,267 @@
...
+from contextlib import asynccontextmanager
+from logging import getLogger
+from workflows.events import StartEvent, Event
+from workflows import Context
+
+
+logger = getLogger(__name__)
+
+
+class WorkflowClient:
```
adrianlyjak, line 25: I don't see a `/handlers` method added here. Should we include that? Related to the `/result` api, it returns a list of the same data model

src/workflows/client/client.py
AstraBert: I added a `serdes_event` function in the server utils, let me know if it aligns with what you were thinking here

src/workflows/client/client.py
AstraBert: I eliminated kwargs for now (seems like the direction we took with the server too, right?)

PR #110, opened 2025-09-23
server: flattened handler metadata in responses, shared OpenAPI components, and test updates

zhaotai:
Summary
- Add and surface flattened runtime handler metadata across server endpoints (no new endpoints).
- Replace per-endpoint, hand-built dicts with a single source `to_dict()` on the server-side wrapper.
- Consolidate OpenAPI schemas via shared components to remove duplication.
- Update tests to reflect expected behavioral changes.

API changes (intentional)
- Flattened metadata fields are now included in these responses:
  - POST /workflows/{name}/run (200)
  - POST /workflows/{name}/run-nowait (200)
  - GET /results/{handler_id} (200, 202)
  - GET /handlers (200, each item)
- Run-nowait response:
  - status was previously “started”; it now returns status: "running".
- Get result error response:
  - GET /results/{handler_id} now returns a detailed error as text/plain on 500 (instead of a JSON object).

Implementation details
- Introduced server-side wrapper `_WorkflowHandler.to_dict()` returning a typed `HandlerDict`.
- Safe accessors on `_WorkflowHandler`:
  - `status`: guards unfinished futures and maps to "running" | "completed" | "failed".
  - `error`: None when unfinished, stringified exception when failed.
  - `result`: None when unfinished; returns final result when completed.
- Centralized response construction to avoid ad-hoc dict assembly.
- OpenAPI:
  - Added shared components:
    - `components.schemas.Handler`: flattened, reusable handler response shape.
    - `components.schemas.HandlersList`: `{ handlers: Handler[] }`.
  - Endpoint docstrings now reference `$ref` to the above, removing duplication.

Backward compatibility notes
- Run-nowait clients should expect `"status": "running"` instead of `"started"`.
- Consumers of GET /results/{handler_id} on error (500) should parse a text/plain body (detailed error message) rather than a JSON object.

Testing
- Updated server tests to reflect the above behavioral changes.
- All server tests passing: 72 passed locally.

Migration/Deployment
- No persistence or schema migrations needed.
- No new configuration flags introduced.

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -242,8 +242,16 @@ class WorkflowServer:
                                     "enum": ["running", "completed", "failed"],
                                 },
                                 "started_at": {"type": "string", "format": "date-time"},
-                                "updated_at": {"type": "string", "format": "date-time", "nullable": True},
-                                "completed_at": {"type": "string", "format": "date-time", "nullable": True},
+                                "updated_at": {
+                                    "type": "string",
+                                    "format": "date-time",
+                                    "nullable": True,
+                                },
+                                "completed_at": {
+                                    "type": "string",
+                                    "format": "date-time",
+                                    "nullable": True,
+                                },
                                 "error": {"type": "string", "nullable": True},
                                 "result": {"description": "Workflow result value"},
                             },
@@ -969,7 +977,9 @@ class _WorkflowHandler:
             status=self.status,
             started_at=self.started_at.isoformat(),
             updated_at=self.updated_at.isoformat(),
-            completed_at=self.completed_at.isoformat() if self.completed_at is not None else None,
+            completed_at=self.completed_at.isoformat()
+            if self.completed_at is not None
+            else None,
             error=self.error,
             result=self.result,
         )
@@ -1021,6 +1031,7 @@ class _WorkflowHandler:
             else:  # otherwise task completed, so nothing else will be published to the queue
                 queue_get_task.cancel()
                 break
+
 
 @dataclass
 class _NamedWorkflow:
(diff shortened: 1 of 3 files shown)


src/workflows/server/server.py
```diff
@@ -40,12 +41,25 @@
...
 
 
+class HandlerDict(TypedDict):
+    handler_id: str
+    workflow_name: str
+    run_id: str | None  # run_id of the handler, easier for debugging
+    error: str | None
+    result: RunResultT | None
+    status: Status
+    started_at: float
```
adrianlyjak: iso time strings are probably easier to read and less ambiguous as to seconds/millis

src/workflows/server/server.py
```diff
@@ -646,48 +694,13 @@ async def _get_handlers(self, request: Request) -> JSONResponse:
...
-                    status = "failed"
-
-            handler_json = {
-                "handler_id": handler_id,
-                "status": status,
-                "result": result,
-                "error": error,
-            }
-            handlers.append(handler_json)
+        items: list[HandlerDict] = []
```
adrianlyjak: nit: can be one liner: `items = [wrapper.to_dict() for wrapper in self._handlers.values()]`

src/workflows/server/server.py
```diff
@@ -930,6 +957,52 @@ class _WorkflowHandler:
...
 
+    # metadata
+    handler_id: str
+    workflow_name: str
+    started_at: float
+    updated_at: float
+    completed_at: float | None
+
+    def to_dict(self) -> HandlerDict:
+        return {
```
adrianlyjak: nit: for TypedDict, I prefer the class like format:
```py
HandlerDict(
   handler_id=self.handler_id,
# ...
```

PR #113, opened 2025-09-25
able to pass in previous handler ID to continue runs in WorkflowServer

logan-markewich:
Simple change to fetch completed runs from the context store and re-use them on the next run

This should enable flows that our users are more familiar with

This means users can
- list existing handlers
- continue runs from an existing handler

Next PR should add pagination to the list handlers endpoint, and probably some "delete" endpoint on the handlers?

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -374,7 +374,7 @@ class WorkflowServer:
             description: Error running workflow or invalid request body
         """
         workflow = self._extract_workflow(request)
-        context, start_event, run_kwargs = await self._extract_run_params(
+        context, start_event, run_kwargs, handler_id = await self._extract_run_params(
             request, workflow.workflow, workflow.name
         )
 
@@ -384,7 +384,6 @@ class WorkflowServer:
             input_ev = None
 
         try:
-            handler_id = nanoid()
             handler = workflow.workflow.run(
                 ctx=context, start_event=input_ev, **run_kwargs
             )
@@ -533,10 +532,9 @@ class WorkflowServer:
             description: Workflow or handler identifier not found
         """
         workflow = self._extract_workflow(request)
-        context, start_event, run_kwargs = await self._extract_run_params(
+        context, start_event, run_kwargs, handler_id = await self._extract_run_params(
             request, workflow.workflow, workflow.name
         )
-        handler_id = nanoid()
 
         if start_event is not None:
             input_ev = workflow.workflow.start_event_class.model_validate(start_event)
@@ -877,7 +875,8 @@ class WorkflowServer:
 
                 context = Context.from_dict(workflow, persisted_handlers[0].ctx)
 
-            return (context, start_event, run_kwargs)
+            handler_id = handler_id or nanoid()
+            return (context, start_event, run_kwargs, handler_id)
 
         except HTTPException:
             # Re-raise HTTPExceptions as-is (like start_event validation errors)
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -353,6 +353,9 @@ class WorkflowServer:
                   context:
                     type: object
                     description: Serialized workflow Context.
+                  handler_id:
+                    type: string
+                    description: Workflow handler identifier to continue from a previous completed run.
                   kwargs:
                     type: object
                     description: Additional keyword arguments for the workflow.
@@ -366,13 +369,13 @@ class WorkflowServer:
           400:
             description: Invalid start_event payload
           404:
-            description: Workflow not found
+            description: Workflow or handler identifier not found
           500:
             description: Error running workflow or invalid request body
         """
         workflow = self._extract_workflow(request)
         context, start_event, run_kwargs = await self._extract_run_params(
-            request, workflow.workflow
+            request, workflow.workflow, workflow.name
         )
 
         if start_event is not None:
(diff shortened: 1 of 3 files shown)


adrianlyjak:
> Next PR should add pagination to the list handlers endpoint, and probably some "delete" endpoint on the handlers?

💯 - especially the delete handlers (and consequence delete running workflow)
Also later add meta data and search for the handlers

src/workflows/server/server.py
```diff
@@ -855,6 +864,18 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
             context = None
             if context_data:
                 context = Context.from_dict(workflow=workflow, data=context_data)
+            elif handler_id:
```
adrianlyjak, line 865: one thing that we might want to be cautious about is validating that its also not actually currently running to prevent duplicates writing to the same state.

src/workflows/server/server.py
```diff
@@ -855,6 +864,18 @@ async def _extract_run_params(self, request: Request, workflow: Workflow) -> tup
...
+                    HandlerQuery(
+                        handler_id_in=[handler_id],
+                        workflow_name_in=[workflow_name],
+                        status_in=["completed"],
+                    )
+                )
+                if len(persisted_handlers) == 0:
+                    raise HTTPException(detail="Handler not found", status_code=404)
+
+                context = Context.from_dict(workflow, persisted_handlers[0].ctx)
```
adrianlyjak, line 876: just confirming, 'cause I don't know, but the handler id is specifiable via the context, right? Or will this make a new handler id?

tests/server/test_server_persistence.py
```diff
@@ -279,3 +289,69 @@ async def test_workflow_cancelled_after_all_retries_fail(
...
+        # Extract context from first run to verify state
+        first_ctx = first_run_persistent.ctx
+        state_data = first_ctx["state"]["state_data"]
+        assert json.loads(state_data["_data"]["count"]) == 5
+        assert json.loads(state_data["_data"]["run_history"]) == ["run_1_increment_5"]
+
+        # Second run - should continue from first run's context, increment by 3
+        handler_id_2 = "cumulative-run-2"
+        # Create context from first run's persisted state
+        continued_ctx = Context.from_dict(workflow=cumulative_workflow, data=first_ctx)
```
adrianlyjak: this has a different handler id looks like? Also kind of manual set up here in this test. Be nice if the server code was organized in such a way that this could more end to end test its full internal behavior

src/workflows/server/server.py
logan-markewich: It will only pull handlers in the `SUCCESS` state

Since we aren't horizontally scaling yet, this should guarantee that its not running. 

Once we need to scale, imo the state store needs to provide some locking mechanism/we need some orchestrator node

src/workflows/server/server.py
logan-markewich: handler_id is something we invented for the workflow server. 

Its assigned here: https://github.com/run-llama/workflows-py/blob/47d4b31960646b206745f009ea6a424a9a37d270/src/workflows/server/server.py#L384

Or here: https://github.com/run-llama/workflows-py/blob/47d4b31960646b206745f009ea6a424a9a37d270/src/workflows/server/server.py#L533

tests/server/test_server_persistence.py
logan-markewich: oh weird, this test seems wrong 

I just built this test off of the existing tests. There's probably some way to call `server._run_workflow()` directly but constructing the request object seemed annoying

adrianlyjak: there's direct api tests that are sort of usable. I agree constructing the request object seems annoying. wondering if we should consider refactoring to something more typed that it calls after parsing the request object, or just more rely on the api style tests

src/workflows/server/server.py
adrianlyjak: if we're to re-use handlers, I think to be more bulletproof, it still needs to check the in memory handlers. For this scenario
1. ran once, succeeded
2. started again, is in progress
3. another request to start again happens, now double writing on the handler id.

tests/server/test_server_persistence.py
```diff
@@ -279,3 +289,42 @@ async def test_workflow_cancelled_after_all_retries_fail(
...
+            assert persisted.status == "completed"
+
+            # Second run - should start with count=5, increment by 3
+            response2 = await client.post(
+                "/workflows/cumulative/run",
+                json={"handler_id": handler_id, "start_event": {"increment": 3}},
+            )
+            assert response2.status_code == 200
+            resp_data2 = response2.json()
+            assert resp_data2["result"] == "count: 8, runs: 2"
```
adrianlyjak, line 330: should make some assertions about the new handler id

src/workflows/server/server.py
logan-markewich: 3. isn't possible if the run is already in progress?

logan-markewich: Note the `status_in=["completed"],` specifier

adrianlyjak: er, I forgot it will persist before completing 🤦

tests/server/test_server_persistence.py
logan-markewich: Fixed this

src/workflows/server/server.py
logan-markewich: Centralized the ID generation

PR #117, opened 2025-09-30
Add WorkflowServer API to stop workflow handler and cancel tasks

adrianlyjak:
Adds an API to cancel and delete running or completed workflows. Includes a query param as to whether it should just cancel, or fully delete

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -235,7 +235,12 @@ class WorkflowServer:
                                 "run_id": {"type": "string", "nullable": True},
                                 "status": {
                                     "type": "string",
-                                    "enum": ["running", "completed", "failed"],
+                                    "enum": [
+                                        "running",
+                                        "completed",
+                                        "failed",
+                                        "cancelled",
+                                    ],
                                 },
                                 "started_at": {"type": "string", "format": "date-time"},
                                 "updated_at": {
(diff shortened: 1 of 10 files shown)


src/workflows/server/server.py
```diff
@@ -800,6 +805,99 @@ async def _post_event(self, request: Request) -> JSONResponse:
...
+                      type: string
+                      enum: [deleted]
+                  required: [status]
+          404:
+            description: Handler not found
+        """
+        handler_id = request.path_params["handler_id"]
+        # Interpret presence of the flag as true unless explicitly false
+        q = request.query_params
+        raw_param = q.get("remove_from_store")
```
adrianlyjak: @cursor this implementation is pretty weird. Let's just support simple booleans in the query like the other apis. The controller method is convoluted with the multiple delete calls. Clean up the control flow. The pre commit configs right now are passing, but remember to also run pre-commit again to verify your change. They can be run by installing the dependencies (uses `uv sync` normally) and running `uv run pre-commit run -a`. Make sure you test both cases, where the handler is present and where it is no longer present

src/workflows/server/server.py
```diff
@@ -800,6 +798,82 @@ async def _post_event(self, request: Request) -> JSONResponse:
                 detail=f"Error processing request: {e}", status_code=500
             )
 
+    async def _delete_handler(self, request: Request) -> JSONResponse:
```
logan-markewich: This name feels wrong, since it doesn't always delete?

I might suggest a different name. `stop_handler` with a cleanup=True/False param? Something else?

adrianlyjak: @logan-markewich Good callout. Also, thoughts on whether the stopping vs deleting should be entirely separate APIs?

right now

```
DELETE /handlers/xyz?stop_only=true
```

I don't think `cleanup` quite communicates the option. In an earlier iteration it was `remove_from_store` which was a bit of a mouthful. I'm maybe leaning towards something like `purge=true` to notate the hard deletion if a single api?"

In terms of the splitting of the APIs, it might make sense in the future to have a bulk deletion, e.g. delete all workflows that are for `workflow_name=foo`.

So a cancel/stop API, maybe still having a `purge` option
```
POST /handlers/xyz/cancel?purge=true
```
and then maybe later a delete api

```
DELETE /handlers?workflow_name=foo&task_id=123&task_id=345
```

logan-markewich: I like the param `purge=True` yea

The suggested POST and DELETE endpoints make more sense than the current yes. The `POST` would essentially be what we have right now in the PR, and we can add DELETE later?

adrianlyjak: Yeah exactly

logan-markewich (approved):

PR #118, opened 2025-09-30
Add consumer lock to events api

adrianlyjak:
Add a consumer lock to the `/events` API handler to ensure reliable event consumption for a single connection, multiple connections are rejected until the first hangs up.


zhaotai:
Not sure in general if it's a good idea because there could be potential cases where users switch to different client or listen to same events from different clients.

Even though we want to limit only one session at the same time. We probably need to implement a mechanism to switch instead of reject.

But as a workaround, it's fine for now

logan-markewich (approved):

PR #123, opened 2025-10-06
Don't skip validation if start event is empty or missing

adrianlyjak:
There was a bug where start event validation was being skipped if the start event dict was empty, since `{}` is falsy in python


logan-markewich (approved):
makes sense to me!

PR #125, opened 2025-10-07
attach metadata to handlers and allow filtering

logan-markewich:
Initial stab at attaching arbitrary metadata to handlers

- metadata is assigned when calling run()
- if continuing a run/session, it will re-use existing metadata (or overwrite)
- the list handlers endpoint now accepts basic filtering params

Concerns
- this technically expands the db for persisted workflows. Do we need to worry about migrations? (sqlite is fine as is imo)
- is there a better way to expose the filters?

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -1136,6 +1136,13 @@ class WorkflowServer:
                     raise HTTPException(detail="Handler not found", status_code=404)
 
                 context = Context.from_dict(workflow, persisted_handlers[0].ctx)
+                if (
+                    handler_metadata is None
+                    and persisted_handlers[0].handler_metadata_json
+                ):
+                    handler_metadata = json.loads(
+                        persisted_handlers[0].handler_metadata_json
+                    )
 
             handler_id = handler_id or nanoid()
             return (context, start_event, handler_id, handler_metadata)


logan-markewich:
Hmm might need to try harder on tests

src/workflows/server/abstract_workflow_store.py
```diff
@@ -23,6 +23,7 @@ class PersistentHandler(BaseModel):
     workflow_name: str
     status: Status
     ctx: dict[str, Any]
+    handler_metadata_json: str | None = Field(default=None)
```
adrianlyjak, line 26: why the nested json? Can't this just be a `dict[str, Any]` (and default to `{}`)?

src/workflows/server/server.py
```diff
@@ -748,16 +760,129 @@ async def _get_handlers(self, request: Request) -> JSONResponse:
...
-        description: Returns all workflow handlers.
+        description: Returns all workflow handlers with optional filtering.
+        parameters:
+          - in: query
+            name: status
+            schema:
+              type: string
+              enum: [running, completed, failed, cancelled]
+            description: Filter by handler status
+          - in: query
```
adrianlyjak, line 771: might make sense to make this consistent with the agent data query DSL https://developers.llamaindex.ai/python/cloud/llamaagents/agent-data-overview/#filter-dsl

I don't love that DSL, but there's not a lot of great options that don't explode in complexity. This DSL looks actually pretty similar? I would like to eventually move that agent data into the open source llamactl (or here?), so developing it out consistently should make it more re-usable and facilitate easier integration.

One difference here is that one is just using a POST to not have to get query parsing involved.

adrianlyjak: we're already parsing json in the query. Could just continue that pattern with the other dsl?

src/workflows/server/server.py
```diff
@@ -748,16 +760,129 @@ async def _get_handlers(self, request: Request) -> JSONResponse:
...
+
+        metadata_has_keys = None
+        if metadata_has_keys_str:
+            metadata_has_keys = [
+                key.strip() for key in metadata_has_keys_str.split(",")
+            ]
+
+        # Filter handlers
+        items = []
+        for wrapper in self._handlers.values():
```
adrianlyjak, line 842: this isn't really supporting completed workflows 😢 . Generally this querying seems like it should be pushed down into the persistence layer, at least in the long term, and we should push some of the state here in the server to an "in memory" persistence layer

src/workflows/server/sqlite_workflow_store.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, db_path: str) -> None:
     def _init_db(self) -> None:
         conn = sqlite3.connect(self.db_path)
         conn.execute(
-            "CREATE TABLE IF NOT EXISTS handlers (handler_id TEXT PRIMARY KEY, workflow_name TEXT, status TEXT, ctx TEXT)"
+            "CREATE TABLE IF NOT EXISTS handlers (handler_id TEXT PRIMARY KEY, workflow_name TEXT, status TEXT, ctx TEXT, handler_metadata_json TEXT)"
```
adrianlyjak, line 19: yeah, this will need a little mini migration, otherwise existing databases will crash and not support metadata. I like this for a no-dependency sqlite version migration pattern:

https://github.com/run-llama/cloud_llama_deploy/blob/main/packages/llamactl/src/llama_deploy/cli/config/_migrations.py

tests/server/test_server_endpoints.py
```diff
@@ -832,7 +832,10 @@ async def test_post_event_missing_event_data(client: AsyncClient) -> None:
 @pytest.mark.asyncio
```
adrianlyjak, line 832: seems like we need a test for querying?

src/workflows/server/abstract_workflow_store.py
logan-markewich: Mostly ease of storage lol but yes, good point

src/workflows/server/server.py
logan-markewich: Oh dang, yea, once workflows complete they get removed from this list

zhaotai:
Maybe this question looks dumb, but why do we need to allow `metadata` to be arbitrary JSON?
Could you help me understand the thinking behind this design?

PR #126, opened 2025-10-07
Expose workflow events in server and workflow class

logan-markewich:
- adds a `workflow.events` property to return a list of all known event types
- exposes `/workflows/{name}/events` in `WorkflowServer` to get event schemas
- adds `additional_events` param to `server.add_workflow()` for cases where events are needed beyond what type annotations expose (like special events needed for `ctx.wait_for_event()`)

The motivation for this is to enhance UI's like the workflow debugger, which would benefit from these schemas when exposing a `send_event()` UI component

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -402,6 +402,7 @@ class WorkflowServer:
         additional_events = self._additional_events.get(name, [])
         if additional_events:
             events.extend(additional_events)
+
         event_objs = []
         for event in events:
             event_objs.append(event.model_json_schema())
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -73,6 +73,7 @@ class WorkflowServer:
         persistence_backoff: list[float] = [0.5, 3],
     ):
         self._workflows: dict[str, Workflow] = {}
+        self._additional_events: dict[str, list[type[Event]] | None] = {}
         self._contexts: dict[str, Context] = {}
         self._handlers: dict[str, _WorkflowHandler] = {}
         self._results: dict[str, StopEvent] = {}
@@ -147,6 +148,11 @@ class WorkflowServer:
                 self._get_workflow_representation,
                 methods=["GET"],
             ),
+            Route(
+                "/workflows/{name}/events",
+                self._list_workflow_events,
+                methods=["GET"],
+            ),
         ]
 
         @asynccontextmanager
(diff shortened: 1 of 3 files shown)


logan-markewich:
Will add one more test tomorrow

src/workflows/server/server.py
```diff
@@ -346,6 +359,56 @@ async def _list_workflows(self, request: Request) -> JSONResponse:
...
+            raise HTTPException(status_code=404, detail=f"Workflow '{name}' not found")
+
+        events = self._workflows[name].events
+        additional_events = self._additional_events.get(name, [])
+        if additional_events:
+            events.extend(additional_events)
+
+        event_objs = []
+        for event in events:
+            event_objs.append(event.model_json_schema())
```
adrianlyjak, line 408: don't we need the fully qualified module name to do anything useful with these for send_event?

logan-markewich: I was thinking to smooth out that api -- if we have a list of events in the server, we can just match based on the schema/names without needing the "qualified" name

adrianlyjak: oh, so like try all schemas and use the first match? That would on the one hand be really nice when it works. There's lots of danger with that approach thought:

For example, this event will parse from anything since it allows extra keys:
```
class Foo(Event):
  pass
```

Even if that's solved, you can also have ambiguity. `{}` could be either `Foo` or `Bar`:
```
class Bar(Event):
  pass
```

The names do have some strong meanings, especially since events are the main dispatch method within workflows, they tend to be small and have meaning in their names.

logan-markewich: No not try all the schemas -- we have a list of the actual event types in memory. So a user can send a more straightforward JSON request

```
{
  "name": "Foo".
  "payload": {}
}
```

Then we can check `event.__name__` against the name in the request to `send_event()`

logan-markewich: This is opposed to the current required schema which is pretty non-user-friendly

I can implement a version of this in this PR too, but wanted to keep these changes minimal

adrianlyjak: oh, yeah, that would be amazing

adrianlyjak: But we still need the name field returned from this API now, right? Otherwise it will be a breaking change to API response to introduce it later

logan-markewich: technically the name is the `title` field from the json schema? But I can make this more obvious

adrianlyjak: Oh, that makes sense

adrianlyjak (approved):

PR #127, opened 2025-10-07
Adrian/context refact

adrianlyjak:


src/workflows/context/context.py
```diff
@@ -115,24 +119,83 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
         # Global data storage
         self._lock = asyncio.Lock()
-        self._state_store: InMemoryStateStore[MODEL_T] | None = None
-        self._waiting_ids: set[str] = set()
 
+        # Init broker machinery
+        self._init_broker_data(previous_context, serializer)
         # instrumentation
         self._dispatcher = workflow._dispatcher
+        self.workflow = workflow
```
logan-markewich: heads up -- this creates a circular reference and might lead to memory not being freed (well, maybe)

Past issue: https://github.com/run-llama/llama_index/issues/18107

src/workflows/workflow.py
```diff
@@ -226,47 +226,18 @@ def _start(
...
+        ctx = (
+            ctx
+            if ctx is not None
+            else Context(
+                self,
+                serializer=ctx._serializer if ctx is not None else None,
+            )
+        )
+        ctx._init_runtime(run_id)
+        self._contexts.add(ctx)
```
logan-markewich: Is it weird to have a list of contexts on the workflow itself? I actually didn't realize this was here

adrianlyjak: purged this, it looks like it was just a kinda funny pattern to obtain the last created context for tests?

src/workflows/context/context.py
adrianlyjak: k, I don't quite understand what happened there, but I changed to just passed the workflow by ref, its sorta cleaner anyway

src/workflows/context/context.py
```diff
@@ -311,138 +344,21 @@ def from_dict(
...
-            # restore waiting ids for hitl
-            context._waiting_ids = set(data["waiting_ids"])
-
-            return context
+            return cls(workflow, previous_context=data, serializer=serializer)
         except KeyError as e:
             msg = "Error creating a Context instance: the provided payload has a wrong or old format."
             raise ContextSerdeError(msg) from e
 
-    async def mark_in_progress(self, name: str, ev: Event, worker_id: str = "") -> None:
```
adrianlyjak, line 361: these weren't prefixed, but they seem very private, so went ahead with deleting them

src/workflows/context/context.py
```diff
@@ -668,263 +508,51 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
-            raise WorkflowCancelledByUser
-        except asyncio.CancelledError:
-            return
+        if self._broker_run is None:
+            raise WorkflowRuntimeError("Workflow run is not yet running")
+        return self._broker_run.streaming_queue
+
+
+@functools.lru_cache(maxsize=1)
+def _warn_get_result() -> None:
```
adrianlyjak, line 591: this seemed a little odd to access this on the context, rather than the handler, which more represents the run (and is literally a future that resolves to this value). I only saw one small test reference.

src/workflows/runtime/broker.py
```diff
@@ -0,0 +1,713 @@
...
+
+    def start(
+        self,
+        workflow: Workflow,
+        start_event: Event | None = None,
+        before_start: Callable[[], Awaitable[None]] | None = None,
+        after_complete: Callable[[], Awaitable[None]] | None = None,
+    ) -> WorkflowHandler:
+        from workflows.handler import WorkflowHandler
+        from workflows.context.context import Context
```
adrianlyjak: circular imports 😞

src/workflows/context/context.py
```diff
@@ -68,6 +65,8 @@ class Context(Generic[MODEL_T]):
     Args:
         workflow (Workflow): The owning workflow instance. Used to infer
             step configuration and instrumentation.
+        previous_context: A previous context snapshot to resume from.
```
logan-markewich, line 74: Why "resume from a previous context" ? Wouldn't you just continue to use the same context object?

src/workflows/context/context.py
```diff
@@ -68,6 +65,8 @@ class Context(Generic[MODEL_T]):
...
+        serializer: A serializer to use for serializing and deserializing the current and previous context snapshots.
 
     Attributes:
         is_running (bool): Whether the workflow is currently running.
```
logan-markewich, line 78: Does the rest of the docstring need to be updated? (attributes, examples, links to relevant API docs?)

src/workflows/context/context.py
```diff
@@ -116,39 +115,141 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+        if self._broker_run is None:
+            return self._init_snapshot.is_running
+        else:
+            return self._broker_run.is_running
 
-        # First-time initialisation
-        self._state_store = InMemoryStateStore(state_class)
+    def _init_broker(self, workflow: Workflow) -> WorkflowBroker[MODEL_T]:
+        if self._broker_run is not None:
+            raise WorkflowRuntimeError("Broker already initialized")
```
logan-markewich, line 214: How would one get to this state?

src/workflows/context/context.py
```diff
@@ -116,39 +115,141 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+        run_id = str(uuid.uuid4())
+
+        broker_state = WorkflowBrokerState.from_serialized(
+            self._init_snapshot, self._serializer
+        )
+        self._broker_run = WorkflowBroker(
```
logan-markewich: Trying to picture some of the hierarchy here
- Context
  - state_store
  - broker_run
    - broker_state

src/workflows/context/context.py
```diff
@@ -116,39 +115,141 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+            workflow=workflow,
+            context=self,
```
logan-markewich, line 220: is it weird to pass in the entire context here? Circular dependency?

src/workflows/context/context.py
```diff
@@ -116,39 +115,141 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+        """
+        prev_broker: WorkflowBroker[MODEL_T] | None = None
+        if self._broker_run is not None:
+            prev_broker = self._broker_run
+            self._broker_run = None
+
+        self._broker_run = self._init_broker(workflow)
+
+        async def before_start() -> None:
+            if prev_broker is not None:
```
logan-markewich, line 243: Didn't we just set it to `None` and re-initialize it above?

src/workflows/context/context.py
```diff
@@ -116,39 +115,141 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+            state=broker_state,
+            run_id=run_id,
+        )
+        return self._broker_run
+
+    def _workflow_run(
+        self,
+        workflow: Workflow,
+        start_event: StartEvent | None = None,
+        semaphore: asyncio.Semaphore | None = None,
```
logan-markewich, line 230: What is the semaphore for?

src/workflows/context/context.py
```diff
@@ -116,39 +115,141 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+            start_event=start_event,
+            before_start=before_start,
+            after_complete=after_complete,
+        )
+
+    def _workflow_cancel_run(self) -> None:
+        """
+        Called internally from the handler to cancel a context's run
+        """
+        if self._broker_run is None:
```
logan-markewich, line 274: this pattern of broker_run being `None` or not seems a little scary 👀

logan-markewich: I might move it to a property to avoid all these duplicate if-checks at the very least

src/workflows/context/context_types.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Events that were in-flight for each step at serialization time. Maps step_name -> [serialized_event_str, ...].
+    in_progress: dict[str, list[str]] = Field(default_factory=dict)
+
+    # Pairs recorded when a step produced an output event: (step_name, input_event_class_name).
+    # Note: stored as Python tuples here; if JSON-encoded externally they become 2-element lists.
+    accepted_events: list[tuple[str, str]] = Field(default_factory=list)
+
+    # Broker log of all dispatched events in order, as serializer-encoded strings.
+    broker_log: list[str] = Field(default_factory=list)
```
logan-markewich, line 46: tbh this isn't used and just eats memory, we could probably delete if we are changing this much

src/workflows/runtime/broker.py
logan-markewich: Context is unused here :)

src/workflows/context/context.py
adrianlyjak: Public API is intended to be the same, other than some methods that appeared to be internal, but were not underscore prefixed

src/workflows/context/context.py
adrianlyjak: Just from tests that are reaching into private methods

src/workflows/context/context.py
adrianlyjak: They are certainly coupled, but have the same lifetime. The context is more or less just the public interface for the broker

src/workflows/context/context.py
adrianlyjak: This was pre-existing, but used to be managed within the workflow. There's an optional max concurrent configuration on the workflow

src/workflows/context/context.py
adrianlyjak: this is the one that was maybe just noned out if pre-existing. Here, the shutdown is within the before_start so that it can be awaited before initializing the new broker.

FWIW, This would only happen if you re-use a context across multiple workflow runs

src/workflows/context/context.py
adrianlyjak: yeah, I can at least add the property. Ideally, the context would initialize the broker_run at initialization, which would make this moot, but it broke a bunch of tests that complained that there is no running loop, so seemed like potentially a bit of a breaking change

src/workflows/context/context_types.py
adrianlyjak: I'm planning more changes, so this seemed like a good checkpoint for review at least to see how we feel about the direction. I think perhaps we should keep these changes on a branch to introduce the full change set in one go?

adrianlyjak:
@logan-markewich 
> This is a decent cleanup. Would probably benefit from an architecture diagram to clarify the current shape and if it matches the desired end-goal
> 
> Also before we merge this (or before we release this), will need to update the API reference docs

Here's a quicky. Basically just split the context into a few separate components that are private implementation details. but let's discuss long term goals more next week. Idea being to be able to sort of switch out the broker/runtime thing with plugins, but there's a number of details in there that we'd still want to re-use across plugins

image

Good point about the reference docs. Those need some attention anyways (they still have stepwise! the errors page is empty?). For the most part the goal here was to make no api surface changes. I'd be inclined to take out Context.__init__ from the docs. Seems like you should only be calling `.from_dict` if manually constructing a context.

src/workflows/context/context.py
adrianlyjak: This is now how you construct from a dict (rather than initializing within the `from_dict` class method)

logan-markewich: why the pattern change though?

adrianlyjak: The previous was setting private instance vars in the classmethod. This now defers the deserialization until the broker starts

src/workflows/context/context.py
```diff
@@ -116,39 +121,161 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+            self._broker_run = None
+
+        self._broker_run = self._init_broker(workflow)
 
-        # First-time initialisation
-        self._state_store = InMemoryStateStore(state_class)
+        async def before_start() -> None:
+            if prev_broker is not None:
+                try:
+                    await prev_broker.shutdown()
```
logan-markewich, line 245: Shouldn't the prev_broker run shutdown before getting to this point? The lifecycle of the broker feels a little odd here

src/workflows/runtime/types/internal_state.py
```diff
@@ -0,0 +1,307 @@
...
+        # Start with a base state from the workflow
+        base_state = BrokerState.from_workflow(workflow)
+        # Always set is_running to False on deserialization - the workflow will set it to True when it starts
+        base_state.is_running = False
+
+        # Restore worker state (queues, collected events, waiters)
+        # We do this regardless of is_running state so workflows can resume from where they left off
+        for step_name, worker_data in serialized.workers.items():
+            if step_name not in base_state.workers:
+                continue
```
logan-markewich, line 146: Isn't it a critical error here? To me this indicates that the context and workflow do not match/are out of sync?

src/workflows/runtime/broker.py
```diff
@@ -0,0 +1,333 @@
...
+    ) -> WorkflowHandler:
+        """Start the workflow run. Can only be called once."""
+        if self._handler is not None:
+            raise WorkflowRuntimeError(
+                "this WorkflowBroker already run or running. Cannot start again."
+            )
+        self._init_state = previous
+
+        async def _run_workflow() -> None:
+            from workflows.context.context import Context
```
logan-markewich: nit: unused import

src/workflows/runtime/broker.py
```diff
@@ -0,0 +1,333 @@
...
+
+    @property
+    def _state(self) -> BrokerState:
+        ticks = self._replay_ticks
+        state = self._init_state or BrokerState.from_workflow(self._workflow)
+        new_state = rebuild_state_from_ticks(state, ticks)
+        return new_state
+
+    @property
+    def _replay_ticks(self) -> list[WorkflowTick]:
```
logan-markewich: I'm not sure if its every commented/explained why we need/want replay? Why did I need to replay ticks?

src/workflows/runtime/control_loop.py
```diff
@@ -0,0 +1,686 @@
...
+        self.state = init_state
+        self.workers: list[asyncio.Task] = []
+        self.queue: asyncio.Queue[WorkflowTick] = asyncio.Queue()
+        self.snapshot_plugin = as_snapshottable(plugin)
+
+    async def wait_for_tick(self) -> WorkflowTick:
+        """Wait for the next tick from the internal queue."""
+        return await self.queue.get()
+
+    def queue_event(self, tick: WorkflowTick, delay: float | None = None) -> None:
```
logan-markewich: nit: `queue_tick` ?

src/workflows/runtime/broker.py
adrianlyjak: Sadly, this needs to be imported if not type checking, and inline for circular import reasons.

src/workflows/runtime/types/results.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+
+    def __init__(self, add: AddWaiter[EventType]):
+        self.add = add
+        super().__init__(f"Waiting for event {add.event_type}")
+
+    add: AddWaiter[EventType]
+
+
+StepWorkerStateContextVar = ContextVar[StepWorkerContext]("step_worker")
```
logan-markewich, line 95: Context vars might not work in a distributed setup? Not a blocker but seems notable

src/workflows/runtime/types/commands.py
```diff
@@ -0,0 +1,77 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Union
+
+from workflows.events import Event, StopEvent
+
+
+@dataclass(frozen=True)
```
logan-markewich, line 25: ooc why dataclass? I know weird things can happen when you mix dataclasses and pydantic, typically I'd recommend picking one or the other

src/workflows/runtime/broker.py
logan-markewich: Oh weird that my ide said it was unused. Makes sense tho

src/workflows/runtime/types/results.py
adrianlyjak: idea would be to set this up so that it works in a distributed context

src/workflows/runtime/types/commands.py
adrianlyjak: they're not meant to be mixed right now, and these are more or less transient. Just wanted something light weight. If we want them to be serializable, should switch to pydantic.

src/workflows/runtime/broker.py
adrianlyjak: Replay is a bad name. This is more or less an event source architecture, so this is basically all you need to record to recreate the state. Right now it's an implementation detail from which we can re-derive the current state rather than syncing the full state from inside the control loop

src/workflows/runtime/types/internal_state.py
adrianlyjak: since this is from serialized state, if you renamed / removed a step, this could happen. Debatable, but seems like it could easily happen in dev. Nice to still be able to best effort deserialize

src/workflows/context/context.py
adrianlyjak: yeah, it is, but you still want the reference to the broker for getting the state, so just double checking here. This also just protects from someone doing weird things with capturing a context that's still running and passing it in again to the workflow.

src/workflows/context/context.py
```diff
@@ -615,43 +513,9 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
-                    continue
-            self.write_event_to_stream(
-                EventsQueueChanged(
-                    name=waiter_id,
-                    size=self._queues[waiter_id].qsize(),
-                )
-            )
+        return await self._running_broker.wait_for_event(
+            event_type, waiter_event, waiter_id, requirements, timeout
+        )
```
dkuettel, line 518: Im not sure if I tracked this down to the right PR or not, but here we go: This seems a pretty hefty change for what came out as a minor version bump (2.8 to 2.9) or so. With this change, from what I understand, `wait_for_event` now raises an exception, and reruns the full step later, whereas before it would block and return the result as soon as it is ready.

Let me know if I should discuss this in another change set.

The problems I see with it:

- the event stream now emits events twice for whatever step is later replayed (yes, because it is in fact run twice), and when using `Workflow.stream_events()` this breaks things, as you dont really expect that, and at least semantically, it is running only once
- people dont know that their steps now have to be "some kind of idempotent"
- And maybe: it's not quite clear to me if this is robust if the second execution changes something that makes the wait for event slightly different, would it recognize that we are waiting for the event from the previously stopped step execution?

Was the previously "plain waiter behavior" bugged, or why was it removed?

adrianlyjak: Hi @dkuettel yes, its an unfortunate side effect of the evolution of the framework. The behavior is [noted in the reference docs](https://developers.llamaindex.ai/python/workflows-api-reference/context/#workflows.context.Context.wait_for_event). The recommended approach for waiting is to instead split your waiters via [InputRequiredEvent/HumanResponseEvent](https://developers.llamaindex.ai/python/llamaagents/workflows/human_in_the_loop/). Splitting this way, there is no need to replay to reach the waiting state. The wait_for_event API is largely a byproduct of agent workflows that use a tool call inside of their step.

> 2. people dont know that their steps now have to be "some kind of idempotent"

Generally, as in other workflows engines, steps are expected to be idempotent, similar to how activities in temporalio, or steps in DBOS are expected to be idempotent. They are the granular unit of replayability. This replay behavior already existed somewhat, the previous expectation is that users may serialize their context, and then resume. In order to actually resume, steps must be replayed. The replay is just more apparent now, the framework has no capability to freeze execution of a function, so IMO its better to be up front about this behavior.

> 1. the event stream now emits events twice for whatever step is later replayed (yes, because it is in fact run twice), and when using Workflow.stream_events() this breaks things, as you dont really expect that, and at least semantically, it is running only once

This only emits the request 1x. If not, that's a bug. See also below for more details

> 3. And maybe: it's not quite clear to me if this is robust if the second execution changes something that makes the wait for event slightly different, would it recognize that we are waiting for the event from the previously stopped step execution?

Yes, this needs to be accommodated in your integration code if using wait_for_event. There's 2 ways to configure this: 1. a stable `waiter_id`. The event will only be emitted when this ID changes. [It defaults to the event name and the requirements](https://github.com/run-llama/workflows-py/blob/a15969bdf9a4ec336cbe52fa90a07a7e7c8b39f7/packages/llama-index-workflows/src/workflows/runtime/broker.py#L309) 2. `requirements` may be passed in to specify whether an event is a match for the wait condition. These are key value pairs matched via equality to the event. The requirements are updated to the latest ones set for [each call to wait_for_event](https://github.com/run-llama/workflows-py/blob/a15969bdf9a4ec336cbe52fa90a07a7e7c8b39f7/packages/llama-index-workflows/src/workflows/runtime/control_loop.py#L547)

dkuettel: Thanks for the quick reply. I did see the documentation, but I stumbled because I didn't realize this change from the change notes. Anyway, forgive me for giving my frustration some air here :) I'm not suggesting to change this. I'm not sure what are my options now:

- I could write my own async wait to reproduce the old behavior, unless you are telling me there were more problems with this approach, other than the issue that we cannot persist or freeze the workflow while it is stuck in this place. This particular issue is not a problem in my setup.
- splitting up into two steps looks very bumpy for me, because I have many such `wait_for_event`s and I have to "lift" them all to the typing system because workflows can only route based on that. I think it would require two more event types for each of those.

As for the repeated events, I think I had a misconception here: The routing events do not appear in `WorkflowHandler.stream_events()`, correct? Only things that are sent explicitely using `Context.send_event`? In this case then the problem is with llamaindex and how it handles tool calls. A tool call in llamaindex that uses `wait_for_event` is run again, and the tool call event is emitted again. I guess it is debatable if events should be re-emitted, but it does feel a bit semi-idempotent :).

PR #129, opened 2025-10-08
Simplify schema for send event to handler in WorkflowServer

logan-markewich:


src/workflows/server/server.py
```diff
@@ -845,6 +845,8 @@ async def _post_event(self, request: Request) -> JSONResponse:
                   event:
                     type: string
```
adrianlyjak, line 846: nit: we should probably encode this {type/data} type in the schema, rather than expecting the stupid nested string. Then the frontend generated client can be happy about sending the proper shape and not have to stringify. 

I made a similar change in the run workflow api, but I guess I didn't do the send event api 😢 

This will need the string still supported for backwards compat, otherwise it'll lead to type errors before the llama-ui is updated to send not strings.

I'm happy to take this on later, just noticing

src/workflows/server/server.py
```diff
@@ -897,7 +903,18 @@ async def _post_event(self, request: Request) -> JSONResponse:
             # Deserialize the event
             serializer = JsonSerializer()
             try:
-                event = serializer.deserialize(event_str)
+                event_data = serializer.deserialize(event_str)
+                if isinstance(event_data, dict):
+                    event_title = event_data["type"]
+                    event = events_by_title[event_title].model_validate(
+                        event_data["data"]
+                    )
```
adrianlyjak, line 915: nit: this may turn into some odd errors if they send invalid dicts without doing existence checks on fields

src/workflows/server/server.py
logan-markewich: Let's do this in another PR (sounds a touch breaking/tricky)

src/workflows/server/server.py
logan-markewich: Since this in a try/except a user should get a nice-enough error (KeyError: 'type' or similar)

but, lemme just make it a check ahead of time

PR #133, opened 2025-10-10
add event envelope

zhaotai:
To handle inherit

### DIFF
--- a/src/workflows/server/serialization.py
+++ b/src/workflows/server/serialization.py
@@ -28,11 +28,10 @@ def _mro_names(cls: type) -> list[str] | None:
     # Skip the class itself by starting from the second MRO entry
     for c in cls.mro()[1:]:
         if c.__name__ in built_in_mros:
-            continue
+            break
         try:
             names.append(c.__name__)
         except Exception:
-            # Best effort; skip if class is unusual
             continue
     if not names:
         return None


src/workflows/server/serialization.py
```diff
@@ -0,0 +1,71 @@
...
+    # Start with the existing JSON-serializable structure
+    base = serializer.serialize_value(event)
+    if not isinstance(base, dict):
+        raise TypeError(f"Expected dict, got {type(base)}")
+
+    envelope = EventEnvelope(
+        value=base.get("value", {}),
+        qualified_name=base.get("qualified_name", ""),
+        __is_pydantic=base.get("__is_pydantic", False),
+        mro=_mro_qualified_names(type(event)),
```
adrianlyjak: I think this is too much information, and kind of ugly with exposing lots of details from the python runtime, and could be excessively large. Or do you think there's a good reason for this in the future? (idk, subclassing some kind of a ChatEvent maybe). 

Really there's just 3 semantic types that are built in: StartEvent, Event, and StopEvent. StartEvent will probably never actually come up as a need for being differentiated thought? So mostly its really just StopEvent. 

Anyways, repeating things like "DictLike" and "BaseModel" in here seem like a waste of bytes, and dealing with the fully qualified name is annoying, and I can't imagine a good use case for sharing the same event name within the same worfkow. Could we tighten this up a bit somehow? My preference would be one of

A) just have a singular type
```py
base_type: Literal["StopEvent", "StartEvent", "Event"]
```
B) if we do need more flexibility
```py
supertypes: ["ChatMessage", "Event"]
```

src/workflows/server/serialization.py
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+    # Start with the existing JSON-serializable structure
+    base = serializer.serialize_value(event)
+    if not isinstance(base, dict):
+        raise TypeError(f"Expected dict, got {type(base)}")
+
+    envelope = EventEnvelope(
+        value=base.get("value", {}),
+        qualified_name=base.get("qualified_name", ""),
```
adrianlyjak, line 55: did you see some of the work that @logan-markewich was doing with the send event? He started modelling a simpler interchange format without the fully qualified name based on reflecting the available event types. Could this be made consistent or share implementation with that? I think these are starting to diverge

https://github.com/run-llama/workflows-py/pull/129/files

src/workflows/server/serialization.py
```diff
@@ -0,0 +1,71 @@
...
+        __is_pydantic=base.get("__is_pydantic", False),
+        mro=_mro_qualified_names(type(event)),
+        origin=(
```
adrianlyjak: What's the use case for this? I can't think of one, it seems like it'd mostly be based on the semantics of the specifc event, and not which library the event is from.

src/workflows/server/serialization.py
zhaotai: This is not for our built-in model. Imagine if user-defined event also has inherit structure. This issue still exists.

But I hear you. I think it makes sense to remove Event, BaseModel, object stuffs here

src/workflows/server/serialization.py
zhaotai: Thank you I didn't notice that. I should have. Thank you for reminding.

src/workflows/server/serialization.py
zhaotai: This is to differentiate user-defined events vs built-in events.
The workflow itself will return all kinds of events no matter they're from builtin or user-defined. So having additional field to flag this difference will make life easier. Otherwise we always need to exhaust our built-in types in client side. Once we add more, it's very easy to omit.

adrianlyjak: I still don't quite understand the use case. In many cases its also sort of a mix (should you consider a `CustomStopEvent` as user or library defined?) We're also discussing wanting to keep response models light, so I think we should really have a use case already ready if adding any field. It's easy to add them, but hard to take them away. Perhaps with new things it can be made into a configurable query param to include it rather than bloating the standard response

src/workflows/server/serialization.py
```diff
@@ -0,0 +1,60 @@
...
+    # New metadata
+    type: str
+    types: list[str] | None
+
+
+def _mro_names(cls: type) -> list[str] | None:
+    built_in_mros = ["Event", "DictLikeModel", "BaseModel", "object"]
+    names: list[str] = []
+    # Skip the class itself by starting from the second MRO entry
+    for c in cls.mro()[1:]:
```
adrianlyjak, line 29: nit: I'd just `break` out once reaching one of these bases. No need to look up their hierarchy. I think as I understand the MRO, once reaching one of these you should be all done

src/workflows/server/server.py
```diff
@@ -636,12 +637,14 @@ async def _stream_events(self, request: Request) -> StreamingResponse:
         description: |
           Streams events produced by a workflow execution. Events are emitted as
           newline-delimited JSON by default, or as Server-Sent Events when `sse=true`.
-          Event data is formatted according to llama-index's json serializer. For
-          pydantic serializable python types, it returns:
+          Event data is returned as an envelope that preserves backward-compatible fields
+          and adds metadata for type-safety on the client:
           {
-            "__is_pydantic": True,
+            "__is_pydantic": true,
```
adrianlyjak, line 643: should we get rid of this line entirely? not sure if it might break our TS types, I don't think we actually reference it anywhere though

zhaotai: Should be fine.

src/workflows/server/serialization.py
zhaotai: ty

PR #134, opened 2025-10-10
Add type validators, and return full types to support extensability

adrianlyjak:
Adds TypedDict objects that are shared between client and server, so that the types don't wander and inconsistencies are obvious.

Additonally refactors the client return results to prefer returning the whole object. This makes future changes easy to be non-breaking in the types. For example we are free to add add pagination and pagination fields to the handler list response.

Also somewhat backs out of the serdes. I think that's confusingly typed right now, and our event request/response typing is in a little bit of a flux. Trying to keep things simple for now to avoid implementations wandering. In the future I think we can clean this up a bit more

### DIFF
--- a/src/workflows/client/client.py
+++ b/src/workflows/client/client.py
@@ -87,11 +87,7 @@ class WorkflowClient:
         """
         if start_event is not None:
             try:
-                start_event = (
-                    start_event
-                    if isinstance(start_event, dict)
-                    else JsonSerializer().serialize_value(start_event)
-                )
+                start_event = _serialize_event(start_event)
             except Exception as e:
                 raise ValueError(
                     f"Impossible to serialize the start event because of: {e}"
@@ -237,11 +233,7 @@ class WorkflowClient:
             SendEventResponse: Confirmation of the send operation
         """
         try:
-            serialized_event: dict[str, Any] = (
-                event
-                if isinstance(event, dict)
-                else JsonSerializer().serialize_value(event)
-            )
+            serialized_event: dict[str, Any] = _serialize_event(event)
         except Exception as e:
             raise ValueError(f"Error while serializing the provided event: {e}")
         request_body: dict[str, Any] = {"event": serialized_event}


examples/client/base/workflow_client.py
```diff
@@ -3,35 +3,26 @@
...
     a: int
     b: int
-    operation: Literal["sum", "subtraction"] = Field(default="sum")
-    _function: Callable[[int, int], int] = PrivateAttr(default=lambda a, b: a + b)
-
-    @model_validator(mode="after")
-    def assign_function(self) -> Self:
-        if self.operation == "subtraction":
-            self._function = lambda a, b: a - b
-        return self
```
adrianlyjak, line 21: silly to add this on the client, when its sending to the server to do the math

examples/client/base/workflow_server.py
```diff
@@ -22,24 +22,12 @@ class OutputEvent(StopEvent):
 
 class AddOrSubtractWorkflow(Workflow):
     @step
-    async def first_step(
-        self, ev: InputNumbers, ctx: Context
-    ) -> CalculationEvent | None:
+    async def first_step(self, ev: InputNumbers, ctx: Context) -> OutputEvent | None:
```
adrianlyjak, line 25: simplified this a bit, I think a lot of this wasn't quite relevant to the demonstration. There's a separate human_in_the_loop example for the complex stuff

tests/client/test_client.py
```diff
@@ -13,47 +13,89 @@ def server() -> WorkflowServer:
...
 
-@pytest.mark.asyncio
-async def test_client(server: WorkflowServer) -> None:
+@pytest.fixture()
+def client(server: WorkflowServer) -> WorkflowClient:
     transport = ASGITransport(server.app)
-    client = WorkflowClient(httpx_kwargs={"transport": transport})
+    return WorkflowClient(httpx_kwargs={"transport": transport})
+
+
```
adrianlyjak, line 21: tests split up in this file to get more granular failures and to clarify intention

src/workflows/protocol/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    source: str
+    target: str
+
+
+class WorkflowGraphNodeEdges(TypedDict):
+    nodes: list[WorkflowGraphNode]
+    edges: list[WorkflowGraphEdge]
+
+
+# Pydantic TypeAdapter validators for lightweight runtime validation/casting
```
logan-markewich: Why not use pydantic objects to begin with?

adrianlyjak: Was thinking about that too. I want to sort of share models with the server. Doing pydantic objects in the server too starts to look a lot like re-inventing fastapi, but wouldn't really be that different here

adrianlyjak: I really don't have a strong preference

One thing to sort of like about the typed dicts is that they basically _are_ the serialized format. That's also something to dislike about them though 😄

examples/client/base/workflow_client.py
AstraBert: Lol, I thought I removed it 🙈

src/workflows/client/client.py
```diff
@@ -223,24 +239,26 @@ async def send_event(
...
-            bool: Success status of the send operation
+            SendEventResponse: Confirmation of the send operation
         """
         try:
-            event = serdes_event(event)
+            serialized_event: dict[str, Any] = (
+                event
+                if isinstance(event, dict)
+                else JsonSerializer().serialize_value(event)
+            )
```
AstraBert: Can we use the `_serialize_event` method also here?

src/workflows/protocol/__init__.py
AstraBert: I would be more in favor of data classes, I feel like they are a good middle ground between typed dicts and pydantic

AstraBert: (I mostly used them for my opencode client, they seem to be working well)

tests/client/test_client.py
AstraBert: Thanks 🙏

src/workflows/protocol/__init__.py
adrianlyjak: They don't have a great serialization story. I love them for light in process, but the `asdict` is an extra step that does a lot of obscuring of types, and for validation, I think pydantic will want you to use their dataclass extension. At that point its basically full fledged pydantic

src/workflows/client/client.py
adrianlyjak: oops, good catch! thanks

PR #143, opened 2025-10-15
Refactor to support runtime plugins

adrianlyjak:
## decouple workflow execution logic from runtime

This is the main change. Decouples our runtime logic (previously in the context and then temporarily moved into the broker), from the asyncio in-memory implementation. The core runtime logic is implemented as events that are serially reduced into state, and subsequent "commands" that should be executed. This approach is verbose, but it offers a lot of granular control, ability to reason about state, and easy unit testing. Adds a plugin based runtime to execute the workflow "somewhere"

- Adjusted to_dict serialization to support newer internal state 
- Add runtime plugin abstraction with a default asyncio-based implementation.
- Add Context.stream_events() and a deprecated queue bridge via Context.streaming_queue.

## Simplify internal events; 

remove `EventsQueueChanged` and simplify `StepStateChanged`

- Remove excessive `StepStateChanged` status. Many statuses were firing effectively simultaneously. now just 3: preparing, in_progress, not_in_progress. 
- Update docs and examples to reflect new APIs and internal events.

## "Breaking" changes 

(these are largely debug and internal apis, so I'm calling it ok)

- Remove `EventsQueueChanged`.
- Replace `Context.streaming_queue` usage with `Context.stream_events()`; a deprecation bridge exists but consumers should migrate. This one is not documented so should be minimal
- Simplify `StepState` (no IN_PROGRESS/EXITED) and `worker_id` may be `None` for PREPARING.
- Remove `InMemoryStateStore.to_dict_snapshot` (public API behavior replaced by `Context.to_dict()`).

## LLM Generated detailed summary of changes

  - New runtime/control loop:
    - `src/workflows/runtime/control_loop.py`: Core reducer and runner.
    - `src/workflows/runtime/types/{commands,internal_state,results,step_function,ticks}.py`: Typed building blocks for control loop.
    - `src/workflows/runtime/workflow_registry.py`: Single-registration for plugins per workflow instance.
    - `src/workflows/runtime/basic_runtime.py`: Default `Plugin` + `WorkflowRuntime` (asyncio) with replay snapshots.
  - Context and broker:
    - `src/workflows/context/context.py`: Plugin injection, V0→V1 validation on load, new `stream_events()` and deprecated `streaming_queue` bridging queue per-call.
    - `src/workflows/runtime/broker.py`: Broker delegates to control loop; state derived from ticks via `BrokerState`.
  - Events and internal events:
    - `src/workflows/events.py`: Simplify `StepState` to PREPARING/RUNNING/NOT_RUNNING; remove `EventsQueueChanged`.
    - Emit consistent `StepStateChanged.input_event_name` and `output_event_name`.
  - Serialization:
    - `src/workflows/context/context_types.py`: Add `SerializedContextV0`, `SerializedContext` (V1), conversion helpers; capture per-step queues, in-progress, collected events, waiters.
    - Remove `InMemoryStateStore.to_dict_snapshot` (unused; snapshot is now covered by `Context.to_dict()` + `BrokerState`).
  - Handler/Workflow:
    - `src/workflows/handler.py`: Switch to `ctx.stream_events()`.
    - `src/workflows/workflow.py`: Localized import for `Context` to avoid import cycles.
  - Server:
    - `src/workflows/server/server.py`: Shut down handlers concurrently with `asyncio.gather`.
  - Docs and examples:
    - `docs/.../managing_events.md`: Remove `EventsQueueChanged` references; clarify `StepStateChanged` semantics and examples.
    - `examples/streaming_internal_events.ipynb`: Update code, remove queue-size logging, refresh outputs; pin kernelspec.
  - Tests:
    - Add runtime tests (`tests/runtime`): control loop integration and transformation unit tests.
    - Update existing tests to reflect new state/ticks-based behavior and removed `EventsQueueChanged`.
    - Update tests to use `TickAddEvent` expectations and V1 serialized context.


src/workflows/runtime/broker.py
```diff
@@ -62,143 +64,44 @@ class UnserializableKeyWarning(Warning):
...
-    # Coordination primitive for event stream writes
-    _step_event_written: asyncio.Condition
-    # Counts of active workers per step
-    _currently_running_steps: DefaultDict[str, int]
-
-    # serializable state stuff
-    _state: WorkflowBrokerState
-
-    # run state
+    _plugin: WorkflowRuntime
```
zhaotai: Do you feel confused when assign runtime to plugin? Especially there is a plugin type.

src/workflows/context/context.py
```diff
@@ -120,23 +124,33 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
     # Backing state store; serialized as `state`
     _state_store: InMemoryStateStore[MODEL_T]
     _broker_run: WorkflowBroker[MODEL_T] | None
+    _plugin: Plugin
+    _workflow: Workflow
```
logan-markewich, line 128: If we add this back, would really recommend running a test for memory leaks before merging

zhaotai:
How do we handle indeterministic steps?
I noticed there are retry logic and in snapshot we only pass necessary state to step function.
So in reality, the step could return different result each time, which will impact retry/replay. How do we deal with it? Or we don't care?

adrianlyjak:
@zhaotai 
> How do we handle indeterministic steps?
> I noticed there are retry logic and in snapshot we only pass necessary state to step function.
> So in reality, the step could return different result each time, which will impact retry/replay. How do we deal with it? Or we > don't care?

In most workflow systems, its ok for steps/activities/tasks to be non-deterministic (e.g. return a random number), what's not ok is randomness within the workflow. With our workflow system, the user cannot write the actual workflow, so as long as this control loop logic is actually deterministic, things should be ok.

Step replay is then usually handled via memoization of the steps result. Resuming the workfly, you don't actually re-run the step, you just retrieve its stored result. If a step is interrupted/crashed it's result will not be recorded yet, so at that point you rerun the step to get its result. For retries, the failures are recorded, and it should only execute the remaining retries

src/workflows/runtime/broker.py
adrianlyjak: Agreed, bad names, will improve

src/workflows/context/context.py
adrianlyjak: There was a leak, but not from here 😅 . The registry concept is still pretty messy. Did weird weak ref things as a workaround, which fixes it https://github.com/run-llama/workflows-py/pull/143/commits/4f77e107ed2ef1e0704d46ddbd0f0095097aeb49, will make that nicer in a future revision

adrianlyjak:
> This looks good to me: the main thing I would add tho would be a demo/guide (could be a doc, could be a notebook) that explains how to exploit all these new changes to build workflows better

@AstraBert I documented the Plugin interface better. I think this is more of a contributing thing rather than an end user. Will iterate on this as we add more runtime plugins

PR #150, opened 2025-10-17
Fix workflow client send event

adrianlyjak:
sadly, I made a last minute cleanup change to the workflow client before merging it in, and didn't realize I broke send event functionality.

The client started sending dicts, and the server only accepts stringified events, and there was no unit test coverage.

- Updated the client to send events as serialized strings
- Updated the server to support dicts
- Added test coverage to both
- Add response body to error messages on the client to make debugging easier on client integrations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dev = [
 
 [project]
 name = "llama-index-workflows"
-version = "2.8.1"
+version = "2.8.2"
 description = "An event-driven, async-first, step-based way to control the execution flow of AI applications like Agents."
 readme = "README.md"
 license = "MIT"
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -1,4 +1,5 @@
 import pytest
+import httpx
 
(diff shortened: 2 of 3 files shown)


src/workflows/server/server.py
```diff
@@ -908,7 +908,11 @@ async def _post_event(self, request: Request) -> JSONResponse:
             # Deserialize the event
             serializer = JsonSerializer()
             try:
-                event_data = serializer.deserialize(event_str)
+                event_data = (
+                    serializer.deserialize(event_str)
+                    if isinstance(event_str, str)
+                    else serializer.deserialize_value(event_str)
+                )
```
AstraBert, line 915: I'm a little confused by this: if we assume (and want) the events to be passed on as serialized strings, why don't we just throw an error when they are not?

adrianlyjak: @AstraBert passing as serialized strings in not really ideal (json-in-json 🤢 ). However its the legacy behavior, so I want to maintain compatibility.

PR #152, opened 2025-10-18
Normalize server/client event schema

adrianlyjak:
Adds a more normalized server/client event schema. Bails on using the `JsonSerializer`, it's recursive thing isn't really useful. 
```js
{
   "type": "Foo",
   "value": { "bar": 1 },
    // optionally included when reading. Maintains support for parsing event from this field
    "qualified_name": "package.Foo",
    // always included when reading, not used when sending events
    "types": ["StartEvent"]
}
```

Maintains support for bare (not wrapped) events when posting a run to a workflow

Additionally, maintains support for all of the many various formats that we've accumulated
- specifying events as strings (both bare events for start_event, and wrapped with type data)
- reading based on the `qualified_name` rather than the `type`
- reading from `data` rather than `value` (this one was recently added. I preferred `value` since its been the historical convention via `JsonSerializer`)

**Note, breaking change!**
- Updated the stream events api in the client to return a pydantic model rather than a bare dict. This has only been released for a week, so seems fine to throw it in real quick

### DIFF
--- a/src/workflows/protocol/serializable_events.py
+++ b/src/workflows/protocol/serializable_events.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 import json
-from typing import Any
+from typing import Any, Type
 
 from pydantic import BaseModel, ValidationError, model_validator
 from workflows.context.utils import import_module_from_qualified_name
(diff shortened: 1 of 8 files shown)


logan-markewich:
The client is also undocumented hehe. Lets tell Diego though

src/workflows/protocol/serializable_events.py
```diff
@@ -0,0 +1,188 @@
...
+
+import json
+from typing import Any, Type
+
+from pydantic import BaseModel, ValidationError, model_validator
+from workflows.context.utils import import_module_from_qualified_name
+from workflows.events import Event
+
+
+class EventEnvelopeWithMetadata(BaseModel):
```
logan-markewich, line 14: Couldn't this extend the EventEnvelope class? Or why do we need two classes?

adrianlyjak: They are sort of divergent. The optionality on the type field will get weird I think

PR #153, opened 2025-10-21
handlers list improvement

zhaotai:
* Always return handlers from db so that we can still get full list even after server restarts.
* Move `_stream_events` to `_WorkflowHandler` to centralize persistence.
* Add metadata to persistence data model.

### Remaining issues:
* After complete, handler should be removed from memory
* Events seem not stored in db. Need to persist events as well.

### DIFF
--- a/src/workflows/server/abstract_workflow_store.py
+++ b/src/workflows/server/abstract_workflow_store.py
@@ -41,4 +41,3 @@ class AbstractWorkflowStore(ABC):
 
     @abstractmethod
     async def delete(self, query: HandlerQuery) -> int: ...
-
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -80,7 +80,9 @@ class WorkflowServer:
         self._contexts: dict[str, Context] = {}
         self._handlers: dict[str, _WorkflowHandler] = {}
         self._results: dict[str, RunResultT] = {}
-        self._workflow_store = workflow_store if workflow_store is not None else MemoryWorkflowStore()
+        self._workflow_store = (
+            workflow_store if workflow_store is not None else MemoryWorkflowStore()
+        )
         self._assets_path = Path(__file__).parent / "static"
         self._persistence_backoff = persistence_backoff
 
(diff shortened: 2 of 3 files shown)


src/workflows/server/sqlite_workflow_store.py
```diff
@@ -116,9 +142,17 @@ def add_in_clause(column: str, values: Sequence[str]) -> None:
 
 
 def _row_to_persistent_handler(row: tuple) -> PersistentHandler:
+    from datetime import datetime
```
adrianlyjak: move imports to top of file

adrianlyjak: FYI, added this in my migrations PR targeting your branch here

zhaotai:
> @zhaotai theres 2 things that I think need resolving
> 
> * This will break existing sqlite databases. Needs a migration. I added here if you want to merge this in: #155. I like this as a super simple pattern for sqlite migrations (using in llamactl for its internal persistence)
> * I believe this is going to break the WorkflowServer now if you don't have persistence. Previously, since it depended on the in memory server state, we got away with having an entirely empty implementation for the workflow store by default (see `EmptyWorkflowStore`). However I think we need to have a full featured in memory implementation by default now instead. There's one started, but its just been used for tests so far. We could dust that off, add some tests for it, and call it the default implementation of the store instead. What do you think?

Agree.
Another question is that how does it affect production db? I didn't see production implementation.

adrianlyjak:
> Agree.
> Another question is that how does it affect production db? I didn't see production implementation.

That one is over here for now:
https://github.com/run-llama/cloud_llama_deploy/blob/main/packages/llama-deploy-appserver/src/llama_deploy/appserver/workflow_store/agent_data_store.py

I believe it will just work since its a json blob of the whole persistent handler

zhaotai:
> > Agree.
> > Another question is that how does it affect production db? I didn't see production implementation.
> 
> That one is over here for now: https://github.com/run-llama/cloud_llama_deploy/blob/main/packages/llama-deploy-appserver/src/llama_deploy/appserver/workflow_store/agent_data_store.py
> 
> I believe it will just work since its a json blob of the whole persistent handler

ok, thank you

PR #155, opened 2025-10-21
Add sqlite migrations

adrianlyjak:


zhaotai:
Side question, how can we know that the migration sql can be removed in the future?

PR #157, opened 2025-10-21
DBOS runtime

adrianlyjak:
- Removes the broker
- Splits up the context into three separate components, hidden behind the existing ctx interface
- extends plugins and renames them to runtimes
- still in progress dbos integration as an alternate runtime

---

<a href="https://app.devin.ai/review/run-llama/workflows-py/pull/157">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,6 +28,9 @@ uv run dev -p server -p client
 uv run dev -- -k test_name
 ```
 
+For more advanced scenarios, you can always `cd packages/some-package` and use pytest directly. The dev tool just provides additional package level test parallelism, and more curated cross package test output to avoid context bloat.
+
+
 ### Linting & Formatting
 ```bash
 uv run pre-commit run -a
--- a/packages/llama-agents-dbos/tests/fixtures/workflows/streaming_interrupt.py
+++ b/packages/llama-agents-dbos/tests/fixtures/workflows/streaming_interrupt.py
@@ -1,6 +1,11 @@
 # SPDX-License-Identifier: MIT
 # Copyright (c) 2026 LlamaIndex Inc.
-"""Streaming stress workflow fixture with many concurrent stream writes."""
+"""Streaming workflow for interrupt/resume testing.
+
+Like streaming_stress but collects all WorkDone events before completing,
+ensuring stream events (including the interrupt signal) are visible before
+the workflow can finish.
+"""
 
 from __future__ import annotations
 
@@ -23,42 +28,40 @@ class WorkItem(Event):
 
 class WorkDone(Event):
     item_id: int = Field(default=0)
-    total_processed: int = Field(default=0)
 
 
 class FanOutComplete(Event):
     pass
 
 
-class StreamingStressWorkflow(Workflow):
+class StreamingInterruptWorkflow(Workflow):
     @step
     async def fan_out(self, ctx: Context, ev: StartEvent) -> FanOutComplete:
-        # Fire many stream writes and internal events concurrently
-        # This creates many background tasks that call DBOS operations
         for i in range(15):
             ctx.write_event_to_stream(ProgressEvent(progress=i))
             ctx.send_event(WorkItem(item_id=i))
         print("STEP:fan_out:dispatched_15_items", flush=True)
-        # Write completion signal to stream for interrupt tests
         ctx.write_event_to_stream(ProgressEvent(progress=999))
         return FanOutComplete()
 
     @step(num_workers=4)
     async def process_work(self, ctx: Context, ev: WorkItem) -> WorkDone:
-        # Each worker also writes to stream, creating more concurrent DBOS ops
-        await asyncio.sleep(0.01)  # Small delay to increase interleaving
+        await asyncio.sleep(0.01)
         ctx.write_event_to_stream(ProgressEvent(progress=100 + ev.item_id))
         print(f"STEP:process_work:{ev.item_id}:complete", flush=True)
         return WorkDone(item_id=ev.item_id)
 
     @step
     async def after_fanout(self, ctx: Context, ev: FanOutComplete) -> None:
-        # Consume FanOutComplete, don't trigger anything
         print("STEP:after_fanout:complete", flush=True)
         return None
 
     @step
-    async def collect(self, ctx: Context, ev: WorkDone) -> StopEvent:
-        # First WorkDone ends the workflow
-        print(f"STEP:collect:{ev.item_id}:complete", flush=True)
-        return StopEvent(result={"first_done": ev.item_id})
+    async def collect(self, ctx: Context, ev: WorkDone) -> StopEvent | None:
+        # Wait for all workers to finish, preventing early completion
+        # that could race with stream event consumption
+        results = ctx.collect_events(ev, [WorkDone] * 15)
+        if results is None:
+            return None
+        print("STEP:collect:all_done", flush=True)
+        return StopEvent(result={"collected": len(results)})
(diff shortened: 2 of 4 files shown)


packages/llama-index-workflows/src/workflows/plugins/dbos.py
```diff
@@ -0,0 +1,107 @@
...
+        workflow: Workflow,
+        workflow_function: ControlLoopFunction,
+        steps: dict[str, StepWorkerFunction],
+    ) -> RegisteredWorkflow | None:
+        """
+        Wrap the workflow control loop in a DBOS workflow so ticks are received via DBOS.recv
+        and sent via DBOS.send, enabling durable orchestration.
+        """
+
+        @DBOS.workflow()
```
qianl15: To make recovery fully work, DBOS workflows need to be registered statically upon start up (currently it registers workflows lazily upon the first run). Since Python functions may not always be serializable and DBOS workflow input must be serializable, one workaround is to maintain a {workflow name: function} definition map.

Then, this DBOS plugin statically registers a generic DBOS workflow function, say `_dbos_control_loop`, that takes in a "workflow function name" parameter, and within the control loop function it'll look up the actual workflow function to run.

packages/llama-index-workflows/src/workflows/plugins/dbos.py
```diff
@@ -0,0 +1,107 @@
...
+        async def _dbos_control_loop(
+            start_event: Event | None,
+            init_state: BrokerState | None,
+            run_id: str,
+        ) -> StopEvent:
+            with SetWorkflowID(run_id):
+                return await workflow_function(start_event, init_state, run_id)
+
+        wrapped_steps = {name: DBOS.step()(step) for name, step in steps.items()}
```
qianl15: For steps to execute in distributed environment, they need to be DBOS workflows. Then the main `run` workflow can enqueue each step worker to distribute the work.

packages/llama-index-workflows/src/workflows/plugins/dbos.py
adrianlyjak: when is the deadline for registration with dbos? Is basically it that the workflows must be registered before calling `DBOS.launch`?

packages/llama-index-workflows/src/workflows/plugins/dbos.py
adrianlyjak: The child workflows would need to be registered at startup time too, correct?

packages/llama-index-workflows/src/workflows/plugins/dbos.py
qianl15: Yes, workflows need to be registered before `DBOS.launch` for the recovery thread and queue dispatching thread to find them.

packages/llama-index-workflows/src/workflows/plugins/dbos.py
qianl15: Yeah, all workflows should be registered at startup time.

packages/llama-agents-dbos/src/llama_agents/dbos/journal/task_journal.py
```diff
@@ -0,0 +1,97 @@
...
+
+import asyncio
+from typing import Any
+
+from sqlalchemy.engine import Engine
+
+from .crud import JournalCrud
+
+
+class TaskJournal:
```
logan-markewich, line 15: kind of crazy that we are doing this ourselves lol

Is this because dbos has no replay?

packages/llama-agents-dbos/src/llama_agents/dbos/runtime.py
```diff
@@ -0,0 +1,680 @@
...
+from workflows.workflow import Workflow
+
+try:
+    from dbos import DBOS, SetWorkflowID
+    from dbos._dbos import _get_dbos_instance
+except ImportError as e:
+    # if 3.9, give a detailed error that dbos is not supported on this version of python
+    if sys.version_info.major == 3 and sys.version_info.minor <= 9:
+        raise ImportError(
+            "dbos is not supported on Python 3.9. Please use Python 3.10 or higher."
```
logan-markewich, line 55: this is already in its own package. Lets just change the min python version?

adrianlyjak: this is definitely a kludge. However going to 3.10 will break the workspace, and things are much easier as a workspace, so I want to avoid as long as possible lol

packages/llama-agents-dbos/src/llama_agents/dbos/journal/task_journal.py
adrianlyjak: yes, totally, its unfortunate, and another kludge. Planning on looking deeper into whether this can get better support within DBOS, but seemed better to start with something working

packages/llama-agents-dbos/src/llama_agents/dbos/runtime.py
```diff
@@ -0,0 +1,680 @@
...
+    """
+    if "schema" in config:
+        return config["schema"]
+    is_postgres = engine.dialect.name == "postgresql"
+    return "dbos" if is_postgres else None
+
+
+# Very long timeout for unbounded waits - encourages workflow to sleep.
+# DBOS's default 60s is too short and gets recorded to event logs.
+_UNBOUNDED_WAIT_TIMEOUT_SECONDS = 60 * 60 * 24  # 1 day
```
logan-markewich, line 100: is 1 day crazy or is that just me? (I'm guessing this is used later on for idle detection?)

packages/llama-agents-dbos/src/llama_agents/dbos/runtime.py
```diff
@@ -0,0 +1,680 @@
...
+        # concurrent writes can read the same max(offset) and try to insert
+        # with the same offset, causing UNIQUE constraint violations.
+        async with self._stream_lock_holder.lock:
+            await DBOS.write_stream_async(_IO_STREAM_PUBLISHED_EVENTS_NAME, event)
+
+    async def get_now(self) -> float:
+        return _durable_time()
+
+    async def send_event(self, tick: WorkflowTick) -> None:
+        # Use run_in_executor to escape DBOS step context.
```
logan-markewich, line 478: clever lol

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -319,6 +332,13 @@ async def run(
                 await self.cleanup_tasks()
                 raise
 
+        # Open the gate briefly so rewind workers can start, then close it
+        # before the main loop begins its get_now() calls.
```
logan-markewich: Is this deterministic?

packages/llama-agents-dbos/src/llama_agents/dbos/runtime.py
adrianlyjak: Every time it wakes, it writes to the journal. Doesn't seem very useful to wake up and write if there's nothing to do, we're just waiting on e.g. a hitl event. Unfortunately it consumes an OS thread (8MB memory, etc). Unfortunately DBOS doesn't release for a sleep either. Planning on checking if they have good ideas or a roadmap for this feature

packages/llama-agents-dbos/src/llama_agents/dbos/runtime.py
logan-markewich: lol fair.... maybe time to revisit the 3.10 framework PR now

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
adrianlyjak: I should test if this works without. coding agent was being defensive. It shouldn't really need to be deterministic since this is inherently a snapshot-based resume, not a replay resume

logan-markewich: Gotcha -- yea just depending on asyncio.sleep(0) to immediately return event loop control seemed scary

adrianlyjak: its gone

PR #159, opened 2025-10-23
Powerful event listeners

johnmalek312:
PR allowing event listeners to modify workflow events without race condition.

eg: 
```python
class MyEvent(Event):
    value: int


class MyWorkflow(Workflow):
    @step
    async def start(self, ctx: Context, _ev: StartEvent) -> StopEvent:
        event = MyEvent(value=1)

        await ctx.write_event_to_stream(event, wait=True)

        assert event.value == 100
        return StopEvent(result=event.value)


async def main():
    workflow = MyWorkflow(timeout=10, verbose=False)
    handler = workflow.run()

    async for event in handler.stream_events():
        if isinstance(event, MyEvent):
            time.sleep(3) # doing something here
            event.value = 100

    result = await handler
    assert result == 100
    print("✅ Test passed")
```
backward compatible

### DIFF
--- a/src/workflows/context/context.py
+++ b/src/workflows/context/context.py
@@ -13,6 +13,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    Coroutine,
     DefaultDict,
     Generic,
     Tuple,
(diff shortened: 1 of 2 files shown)


johnmalek312:
requires event listener

PR #171, opened 2025-10-29
Return fully qualified type from handler result

adrianlyjak:
Updates handler APIs to no longer unwrap the `StopEvent.result`. 

Additionally dumps the event as a stop event envelope. 

I'm not actually sure whether it should be a bare event, or the event envelope. I was inclined to make it qualified, since that more aligns with the current react integration's expectations. However the prior knowledge of the `StopEvent` is perhaps similar to that of the `StartEvent`. However, I you technically can have multiple types of `StopEvent`s unlike `StartEvent` (imagine this is rare).

That being said, adding the envelope makes this change even more breaking. I'm second guessing myself here, and am wondering whether we should have an entirely new field or some way to opt in to the change to make it non-breaking

### Custom `MyStopEvent`

given
```py
class MyStopEvent(StopEvent):
    foo: str
MyStopEvent(foo="bar")
```

handler result
```js
// before
{
  // ...
  "result": { "foo": "bar" }
}
// this change
{
  // ...
  "result": { 
    "type": "MyStopEvent",
    "qualified_name": "my_package.MyStopEvent",
    "types": ["StopEvent"],
    "value": { "foo": "bar" }
  }
}

```

### Plain `StopEvent`

given
```py
StopEvent(foo="bar", result="this is the embedded result")
```
handler result before
```js
// before
{
  // ...
  "result": "this is the embedded result
}
// this change
{
  // ...
  "result": { 
    "type": "StopEvent",
    "qualified_name": "workflows.events.StopEvent",
    "value": { "foo": "bar", "result": "this is the embedded result"}
  }
}
// maybe we should do this instead to keep it less breaking and simple?
{
  // ...
  "result":  { "foo": "bar", "result": "this is the embedded result"}
}
```

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -709,9 +709,9 @@ class WorkflowServer:
             existing = found[0]
             return _WorkflowHandler.handler_data_from_persistent(existing)
         else:
-            if wrapper.run_handler.done():
+            if wrapper.run_handler.done() and wrapper.task is not None:
                 try:
-                    await wrapper.run_handler  # make sure its fully done
+                    await wrapper.task  # make sure its fully done
                 except Exception:
                     # failed workflows raise their exception here
                     pass  # failed workflows raise their exception here


zhaotai:
I'm inclined to have it qualified because in FE we can process all events using the same processor.

In terms of breaking changes,
I encountered this issue in llama-ui as well. But then I realized that it's better to make breaking changes now while there are fewer users than later when there are more. Also, constantly worrying about breaking changes can significantly slow us down and prevent us from making the best decisions. I think we should just change the version number to alpha and freely make breaking changes.

src/workflows/server/server.py
```diff
@@ -693,31 +698,53 @@ async def _run_workflow_nowait(self, request: Request) -> JSONResponse:
...
+                HandlerQuery(handler_id_in=[handler_id])
+            )
+            if not found:
+                raise HTTPException(detail="Handler not found", status_code=404)
+            existing = found[0]
+            return _WorkflowHandler.handler_data_from_persistent(existing)
+        else:
+            if wrapper.run_handler.done():
+                try:
+                    await wrapper.run_handler  # make sure its fully done
```
zhaotai: await wrapper.task

PR #172, opened 2025-10-29
docs: github action and pkg docs

AstraBert:
Adding docs for Github Action (they were not included when we migrated the documentation here) and docs for the `pkg` command

### DIFF
--- a/docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
+++ b/docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
@@ -5,16 +5,12 @@ sidebar:
 ---
 
 :::caution
-This command is currently limited to **agent workflows**.
+This command is currently limited to python **agent workflows**.
 Frontend packaging is **not yet supported**.
 :::
 
-The `pkg` command lets you package your application in different formats.
-Currently supported formats: **Docker** and **Podman**.
-
-It generates minimal, customizable build files from a deployment configuration, enabling you to easily package and deploy your application using container platforms.
-
-For details on configuring your deployment file, see the [Deployment Config Reference](/python/cloud/llamaagents/configuration-reference).
+The `pkg` command group lets you package and export your application for custom deployments.
+Currently it supports exporting a Dockerfile that can be built into an image with any OCI compliant image builder, such as `docker` or `podman`.
 
 
 ## Usage
@@ -25,7 +21,7 @@ llamactl pkg [COMMAND] [options]
 
 ### Available Commands
 
-- `container` – Generate a minimal, build-ready container file (e.g., `Dockerfile`) for your workflows using Docker or Podman.
+- `container` – Generate a minimal, build-ready container file (e.g., `Dockerfile`) for your workflows.
 
 
 ## Command: `container`
@@ -34,7 +30,7 @@ llamactl pkg [COMMAND] [options]
 llamactl pkg container [DEPLOYMENT_FILE] [options]
 ```
 
-Generates a container build file compatible with **Docker** and **Podman** from a given deployment file and user-specified options.
+Generates a container build file from a given deployment file and user-specified options.
 
 ### Options
 
--- a/docs/src/content/docs/llamaagents/llamactl/getting-started.md
+++ b/docs/src/content/docs/llamaagents/llamactl/getting-started.md
@@ -135,14 +135,9 @@ After creation, the TUI will show deployment status and logs.
 
 ## Alternative Ways to Deploy
 
-If you prefer a self-hosted deployments, `llamactl` provides the possibility of exporting your workflows as minimal and customizable container files compatible with Docker and Podman, so that you can easily containerize and deploy your agent workflow anywhere.
+If you prefer a self-hosted deployments, `llamactl` has utilities to export your workflows to container files, so that you can easily build images and deploy your agent workflow anywhere.
 
 Read more about this in the [`pkg` command reference](/python/cloud/llamaagents/llamactl-reference/commands-pkg).
-
-:::caution
-This command is currently limited to **agent workflows**.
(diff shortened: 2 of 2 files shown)


docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
```diff
@@ -0,0 +1,74 @@
+---
+title: pkg
+sidebar:
+  order: 106
+---
+
+:::caution
+This command is currently limited to **agent workflows**.
```
adrianlyjak: I think this isn’t very clear to a newcomer what it means. Maybe just add the word “python” would help?

docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
```diff
@@ -0,0 +1,74 @@
...
+Frontend packaging is **not yet supported**.
+:::
+
+The `pkg` command lets you package your application in different formats.
```
adrianlyjak: I don't know if we really want to go hard on lots of different formats, so maybe let's not sell it that way. Docker/OCI basically covers 99% of packaging use cases.

Suggest instead
```
The `pkg` command group lets you package and export your application for custom deployments.
Currently it supports exporting a Dockerfile that can be built into an image with any OCI compliant image builder, such as `docker` or `podman`.
```

docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
```diff
@@ -0,0 +1,74 @@
...
+Currently supported formats: **Docker** and **Podman**.
+
+It generates minimal, customizable build files from a deployment configuration, enabling you to easily package and deploy your application using container platforms.
+
+For details on configuring your deployment file, see the [Deployment Config Reference](/python/cloud/llamaagents/configuration-reference).
+
```
adrianlyjak: nit maybe delete. Seems unnecessary. 

```suggestion
```

docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
```diff
@@ -0,0 +1,74 @@
...
+## Usage
+
+```bash
+llamactl pkg [COMMAND] [options]
+```
+
+### Available Commands
+
+- `container` – Generate a minimal, build-ready container file (e.g., `Dockerfile`) for your workflows using Docker or Podman.
```
adrianlyjak:
```suggestion
- `container` – Generate a minimal, build-ready container file (e.g., `Dockerfile`) for your workflows.
```

docs/src/content/docs/llamaagents/llamactl-reference/commands-pkg.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+## Command: `container`
+
+```bash
+llamactl pkg container [DEPLOYMENT_FILE] [options]
+```
+
+Generates a container build file compatible with **Docker** and **Podman** from a given deployment file and user-specified options.
```
adrianlyjak:
```suggestion
Generates a container build file from a given deployment file and user-specified options.
```

docs/src/content/docs/llamaagents/llamactl/getting-started.md
```diff
@@ -131,7 +131,18 @@ When you save, LlamaAgents will verify that it has access to your repository (an
 After creation, the TUI will show deployment status and logs.
 - You can later use `llamactl deployments get` to view again.
 - You can add secrets or change branches with `llamactl deployments edit`.
-- If you update your source repo, run `llamactl deployments update` to roll a new version.
+- If you update your source repo, run `llamactl deployments update` to roll a new version. If you are interested in automating the deployment updates with GitHub Actions, you can read the [Continuous Deployment guide](/python/cloud/llamaagents/cd-with-github-actions)
+
+## Alternative Ways to Deploy
+
+If you prefer a self-hosted deployments, `llamactl` provides the possibility of exporting your workflows as minimal and customizable container files compatible with Docker and Podman, so that you can easily containerize and deploy your agent workflow anywhere.
```
adrianlyjak:
```suggestion
If you prefer a self-hosted deployments, `llamactl` has utilities to export your workflows to container files, so that you can easily build images and deploy your agent workflow anywhere.
```

docs/src/content/docs/llamaagents/llamactl/getting-started.md
```diff
@@ -131,7 +131,18 @@ When you save, LlamaAgents will verify that it has access to your repository (an
...
+
+Read more about this in the [`pkg` command reference](/python/cloud/llamaagents/llamactl-reference/commands-pkg).
+
+:::caution
+This command is currently limited to **agent workflows**.
+Frontend packaging is **not yet supported**.
+:::
```
adrianlyjak: nit: Seems repetitive to warn here too. We can add this soon

```suggestion
```

AstraBert:
> Suggested some changes to keep the docs lean. There's a whole eco system of OCI builders/runtimes etc. so doesn't seem useful to call out Podman over and over, and its a standard format, so IMO seems better to keep general.
> 
> Aside, for later, this kind of calls out a need to provide guidelines on architecture and limitations (e.g. no replicas at the moment, persistence is minimally configurable between sqlite and agent data).

Thanks for the review @adrianlyjak, I applied the changes :)) 

Are you saying that we should provide the guidelines later on because you want to "battle-test" this more and try and fix the limitations?

adrianlyjak:
> Are you saying that we should provide the guidelines later on because you want to "battle-test" this more and try and fix the limitations?

Just seems separate from this change. No need to do everything here

adrianlyjak (approved):

Issue #173, opened 2025-10-29
I encountered an error: cannot pickle 'module' object


lf2foce:
**Issue Description:**
The multiagent workflow breaks when using llama-index-workflows version 2.9.0, while it works perfectly with version 2.8.3.

**Problem Details:**
- Error: `TypeError: cannot pickle 'module' object` during workflow context serialization
- The issue occurs specifically with `deepcopy()` operations in workflow context handling
- This wasn't included in the main llamaindex v0.14.6 release, but when deployed to production, it breaks the entire multiagent system

**Environment:**
- llama-index-workflows: 2.9.0 (broken) vs 2.8.3 (working)
- Python deepcopy fails on workflow context objects containing module references

**Temporary Workaround:**
Downgraded to llama-index-workflows==2.8.3 and pinned the version to prevent auto-updates.

**Request:**
Could you please investigate the serialization changes introduced in version 2.9.0 that might be causing this pickling issue with workflow contexts?

logan-markewich:
@lf2foce any code to reproduce would help

adrianlyjak:
@lf2foce a stack trace would also be helpful

francvi:
I've detected exactly the same with v2.9.0, and downgrading to v2.8.3 solved the issue. I'll try to provide a reproduction code tomorrow if still needed.

Disclaimer: In my case, it didn't raise any error directly, the agent.run() simply stops before really start working.

adrianlyjak:
Thanks @bambooblood. We've also pushed a fix for hashing issue, but it doesn't sound like the same. I'm curious if you see the same issues on v2.9.1

francvi:
Sure, I'll take a look tomorrow!

francvi:
Hey @adrianlyjak ! I've done some testing switching between different versions:

Both v2.8.3 and v2.9.1 are working as expected! Using the provided code (a very simplified version of my current implementation) you should be able to see the events in the console. So, seems like the issue is only happening in v2.9.0, where you'll see that the agent is not really running and it only calls the done_callback without any other kind of interaction. I've taken a look at your PR and I'm afraid I didn't see a direct relationship between it and any of this.

```
# requirements.txt
llama-index == 0.14.6
llama-index-core == 0.14.6
llama-index-llms-azure-openai == 0.4.2
# llama-index-workflows == 2.8.3 # Working as expected
llama-index-workflows == 2.9.0 # Does not work
# llama-index-workflows == 2.9.1 # Working as expected
```

[...]

PR #177, opened 2025-10-29
fix: Fix for unhashable workflow subclasses

adrianlyjak:
The llama_index core workflows are unhashable. The intent here is to just hash on instances anyways. Made a custom weak dict to support this. With this change, and some small mocking changes, llama_index core tests pass again with the new version.

Probably still something else going on with the multi agent serialization errors (#173)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dev = [
 
 [project]
 name = "llama-index-workflows"
-version = "2.9.0"
+version = "2.9.1"
 description = "An event-driven, async-first, step-based way to control the execution flow of AI applications like Agents."
 readme = "README.md"
 license = "MIT"
--- a/uv.lock
+++ b/uv.lock
@@ -1331,7 +1331,7 @@ dev = [
 
 [[package]]
 name = "llama-index-workflows"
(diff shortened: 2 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,76 @@
+# LlamaIndex Workflows - Claude Development Guide
```
adrianlyjak, line 1: this is just an (unrelated) move of claude.md, and adding a final section for "Autonomous Operation". Not sure why it doesn't show as a rename. CLAUDE.md is symlinked to AGENTS.md, so should function the same

src/workflows/runtime/types/_identity_weak_ref.py
logan-markewich, line 1: ngl no idea what this file is doing, but I can run with it lol

adrianlyjak: The wrapper is a weak ref that wraps a target object, and uses that objects identity as its hash, so it can go in a dict. In turn it has a de-reference callback that pops it from the parent dict. It's basically a big workaround to use identity as a hash instead of a normal hash for keys.

PR #178, opened 2025-10-29
fix bugs

zhaotai:
1. Fix bug where finished handler won't be removed from memory
2. Fix bug where `run_workflow` didn't await `handler.task`.

### DIFF
--- a/src/workflows/server/sqlite/migrations/0003_add_events_column.sql
+++ b/src/workflows/server/sqlite/migrations/0003_add_events_column.sql
@@ -1,4 +1,0 @@
-PRAGMA user_version=3;
-
--- Add events column to store event history as JSON
-ALTER TABLE handlers ADD COLUMN events TEXT DEFAULT '[]';
--- a/docs/src/content/docs/llamaagents/workflows/deployment.md
+++ b/docs/src/content/docs/llamaagents/workflows/deployment.md
@@ -95,7 +95,7 @@ The `WorkflowServer` exposes the following RESTful endpoints:
 | `GET`  | `/workflows`                   | Lists the names of all registered workflows.                                                            |
 | `POST` | `/workflows/{name}/run`        | Runs the specified workflow synchronously and returns the final result.                                 |
 | `POST` | `/workflows/{name}/run-nowait` | Starts the specified workflow asynchronously and returns a `handler_id`.                                |
-| `GET`  | `/results/{handler_id}`        | Retrieves the result of an asynchronously run workflow. Returns `202 Accepted` if still running.       |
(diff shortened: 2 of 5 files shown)


src/workflows/runtime/broker.py
```diff
@@ -162,11 +161,7 @@ async def _run_workflow() -> None:
                     finally:
                         # ensure run context is cleaned up even on failure
                         workflow_registry.delete_run(run_id)
-                    result.set_result(
-                        workflow_result.result
-                        if type(workflow_result) is StopEvent
-                        else workflow_result
-                    )
+                    result.set_result(workflow_result)
```
adrianlyjak: No, this is a big breaking change. It sets the result of the `WorkflowHandler` future. The value here is the result of `await workflow.run()`. That's for the next major version and needs to be coordinated with framework codebase

src/workflows/protocol/__init__.py
```diff
@@ -34,6 +34,12 @@ class WorkflowsListResponse(BaseModel):
     workflows: list[str]
 
 
+class WorkflowResultResponse(BaseModel):
```
adrianlyjak: why not return the handler object? Isn't this effectively a subset?

src/workflows/server/server.py
```diff
@@ -677,46 +691,74 @@ async def _get_workflow_result(self, request: Request) -> JSONResponse:
...
-            description: Error computing result
-            content:
-              text/plain:
-                schema:
-                  type: string
         """
         handler_id = request.path_params["handler_id"]
 
         wrapper = self._handlers.get(handler_id)
-        if wrapper is None:
```
adrianlyjak: Is the reason you added the new api to load it from the db? This just seems like a bug here in the existing handler/result api. We should be loading it from the DB here if not present. Otherwise this endpoint is not very useful

src/workflows/server/sqlite/migrations/0003_add_events_column.sql
```diff
@@ -0,0 +1,4 @@
+PRAGMA user_version=3;
+
+-- Add events column to store event history as JSON
+ALTER TABLE handlers ADD COLUMN events TEXT DEFAULT '[]';
```
adrianlyjak: this looks like an oops?

src/workflows/protocol/__init__.py
zhaotai: I was considering that. But semantically it's different. I'm also thinking of just removing `get_result` and keep `get_handler`.

src/workflows/server/server.py
zhaotai: yep. That's one of the reason.

src/workflows/protocol/__init__.py
adrianlyjak: Agreed that the handler name makes more sense at this point

PR #186, opened 2025-11-02
validate that workflow is not cleared until events are consumed

adrianlyjak:
Events were not retrievable for workflows that complete very fast, the events endpoint was returning a 204, despite there being unconsumed events

### DIFF
--- a/src/workflows/server/server.py
+++ b/src/workflows/server/server.py
@@ -44,6 +44,7 @@ from workflows.protocol import (
     WorkflowEventsListResponse,
     WorkflowGraphResponse,
     WorkflowSchemaResponse,
+    is_status_completed,
 )
 from workflows.server.abstract_workflow_store import (
     AbstractWorkflowStore,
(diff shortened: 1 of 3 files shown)


src/workflows/server/server.py
```diff
@@ -1125,9 +1125,12 @@ async def _post_event(self, request: Request) -> JSONResponse:
 
         # Check if handler exists
         wrapper = self._handlers.get(handler_id)
+        completed_statuses = {"completed", "failed", "cancelled"}
```
logan-markewich: Feels like the status' should be an enum somewhere? But eh

logan-markewich (approved):

src/workflows/server/server.py
adrianlyjak: Good point. there's a literal union. I put a helper function alongside to colocate the check https://github.com/run-llama/workflows-py/pull/186/commits/c974a9c366d77d2b6f8ac01e22feb18b78cb6acd

PR #187, opened 2025-11-02
handle non serializable requirements

adrianlyjak:
related to #173 .

1. Fixes by no longer serializing the requirements, but rather re-initializing the requirements on reload of a state.
2. Additionally, the call to deep copy was hitting the non-serializable object during deep copy, recreating the initial bug report, so added a domain specific deep copy.

Its possible that fix 1 could be achieved by rather serializing the requirements via the serializer? Since these requirements are intended to match against the value of an event, they should also be serializable in a similar manner. Unfortunately the path doesn't go through pydantic serialization, so may be difficult

### DIFF
--- a/tests/test_workflow.py
+++ b/tests/test_workflow.py
@@ -12,6 +12,7 @@ import weakref
 from typing import Any, Callable, Union
 from unittest import mock
 
+from pydantic import PrivateAttr
 import pytest
 
 from workflows.context import Context, PickleSerializer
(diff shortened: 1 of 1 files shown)


src/workflows/runtime/control_loop.py
```diff
@@ -365,7 +366,7 @@ def _process_step_result_tick(
     """
     processes the results from a step function execution
     """
-    state = deepcopy(init)
+    state = init.deepcopy()
```
logan-markewich, line 369: Meta comment, Seems kind of expensive to deepcopy state just to process a single tick 🤔

adrianlyjak: This should be very cheap now with the custom copy implementation here. FWIW, I didn’t see any performance regressions

logan-markewich: Fair!

adrianlyjak: We might want to reconsider the defensive copying here. Initial implementation here is a functional style--returning a new state from the function rather than mutating it, but didn't want to do verbose copies for each update, so just made it one big copy at the start of the tick handler, with mutations within.
This potentially could a) do more granular copying or b) switch to an entirely mutable approach.

PR #194, opened 2025-11-05
Improve logging

adrianlyjak:
- Adds improved error logging to make failures more visible
- Adds additional dispatcher context tags:
  - handler_id from the workflow server
  - run_id from the workflow run itself
- Adds example demonstrating that these tags are available to loggers (struct log, standard lib loggin)
- switches run_id to a nanoid

### DIFF
--- a/src/workflows/runtime/broker.py
+++ b/src/workflows/runtime/broker.py
@@ -125,7 +125,7 @@ class WorkflowBroker(Generic[MODEL_T]):
         self._init_state = previous
 
         async def _run_workflow(run_id: str, tags: dict[str, Any]) -> None:
-            with instrument_tags({**tags, "run_id": run_id}):
+            with instrument_tags({"run_id": run_id, **tags}):
                 # defer execution to make sure the task can be captured and passed
                 # to the handler as async exception, protecting against exceptions from before_start
                 self._is_running = True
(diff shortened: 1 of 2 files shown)


examples/observability/context_logging_structlog.py
logan-markewich, line 1: minor, but I feel like this is pretty complicated/scary to read as an example

logan-markewich: Maybe thats just how structlog works lol but anything to make this more minimal would help

src/workflows/runtime/broker.py
```diff
@@ -119,67 +124,70 @@ def start(
...
-            # to the handler as async exception, protecting against exceptions from before_start
-            self._is_running = True
-            await asyncio.sleep(0)
-            if before_start is not None:
-                await before_start()
-            try:
-                init_state = previous or BrokerState.from_workflow(workflow)
-
+        async def _run_workflow(run_id: str, tags: dict[str, Any]) -> None:
+            with instrument_tags({**tags, "run_id": run_id}):
```
logan-markewich: small chance this conflicts with a users own tags, but eh

adrianlyjak: I suppose I could re-order to prefer theirs

examples/observability/context_logging_structlog.py
adrianlyjak: yeah, fair, I just copy pasted some stuff, I think I could make this less noisy. I could probably also convert to a notebook, and at least that will be celled away

PR #199, opened 2025-11-07
Fix reconnect

adrianlyjak:
Fixes a sad bug where anytime any streaming connection disconnects, it discards the entire workflow from memory, making reconnects impossible, and leaving the workflow in a dangling "running" state

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dev = [
 
 [project]
 name = "llama-index-workflows"
-version = "2.11.0"
+version = "2.11.1"
 description = "An event-driven, async-first, step-based way to control the execution flow of AI applications like Agents."
 readme = "README.md"
 license = "MIT"
--- a/uv.lock
+++ b/uv.lock
@@ -1560,7 +1560,7 @@ dev = [
 
 [[package]]
 name = "llama-index-workflows"
-version = "2.11.0"
+version = "2.11.1"
 source = { editable = "." }
 dependencies = [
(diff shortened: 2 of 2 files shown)


src/workflows/server/server.py
```diff
@@ -1641,7 +1644,8 @@ async def _iter_events(self, timeout: float = 1) -> AsyncGenerator[Event, None]:
                     queue_get_task.cancel()
                     break
         finally:
```
adrianlyjak, line 1646: this recent addition to better clean up resources. was getting triggered when the api disconnected (disconnecting from the async generator, triggering a cancellation)

logan-markewich (approved):

src/workflows/server/server.py
```diff
@@ -522,6 +522,9 @@ async def _run_workflow(self, request: Request) -> JSONResponse:
                     await wrapper.task
                 except Exception:
                     pass
+            # explicitly close handlers from this synchronous api so they don't linger with events
+            # that no-one is listening for
+            await self._close_handler(wrapper)
```
adrianlyjak, line 527: Also adding behavior for the plain run workflow api to clean itself out of memory upon completion

tests/server/test_server_live_http.py
```diff
@@ -0,0 +1,152 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2025 LlamaIndex Inc.
```
adrianlyjak, line 2: the ASGI test client doesn't handle streaming well. This adds some real (slower) http tests. They're not that slow though

PR #202, opened 2025-11-14
Update workflows for better mono-repo build support

adrianlyjak:
- Moves workflows-py into packages
- Refactors and adds to the workflows scripts, moving them into their own workflows-dev package
- Updates coverage to run against all packages, not just llama-index-workflows
- Updates release process. It's now triggered automatically by bumping the package version and pushing to main

## tagging
PRs are automatically tagged based on changed folders with e.g. `pkg:llama-index-workflows`

## Details of Release process

- Update pyproject(s) versions
- Github action will compare each package's version to the the last highest tag with its prefix
- If increased, it will trigger a release
  - runs the build
  - publishes
  - pushes a prefixed tag, e.g. `llama-index-workflows@v2.11.2`
  - generates release notes by querying github api for the tags, and scooping up any PRs within the range that have a matching tag for the package
  - shoving the PR titles in the release notes. (This still uses the enhancement, bug, breaking-change tags, if applied)
  - triggering the openapi publishing if configured to do so


scripts/generate_release_notes.py
logan-markewich, line 1: Should this be part of workflows-dev or nah?

adrianlyjak: Ah geez, vibe coding gone wrong, that was the intent

adrianlyjak: Yeah this is a stray extra file

.github/release.yml
```diff
@@ -1,14 +1,11 @@
 changelog:
```
adrianlyjak: Oops, I think this should actually be deleted now

PR #207, opened 2025-11-17
Attempting to add changesets

AstraBert:
Attempt to add changesets + related CI actions for versioning and releasing (should be working)


scripts/new_version.py
```diff
@@ -0,0 +1,19 @@
+import toml  # type: ignore [import]
```
adrianlyjak: the purpose of workflows-dev is to centralize these misc scripts in a package. Ideally the changesets code would go there too

packages/workflows-dev/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
adrianlyjak: this should not be published

package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "llama-agents-workspace",
+  "version": "1.0.0",
+  "private": "true",
+  "license": "MIT",
+  "scripts": {
+    "pre-commit-version": "pnpm changeset",
+    "version": "pnpm version-workflows-dev && pnpm version-llama-index-workflows && pnpm version-llama-index-utils-workflow",
+    "release": "pnpm release-workflows-dev && pnpm release-llama-index-workflows && pnpm release-llama-index-utils-workflow",
```
adrianlyjak: this seems rather roundabout. 

1. changesets calls release on the root workspace, 
2. the root workspace calls another script on all packages, manually defined
3. each script calls the individual package script
4. which calls a shared bash script

You could just jump straight from 1 to 4, and have the script iterate through all packages (or have them hard coded)

scripts/requirements.txt
```diff
@@ -0,0 +1,2 @@
+toml
```
adrianlyjak: why are we using a requirements file? We're already using uv for dependency management

scripts/release.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# check pre-release
+uv venv && source .venv/bin/activate && uv pip install -r scripts/requirements.txt
+do_not_release=$(python3 scripts/pre_release_check.py --package $package)
+
+if [ "$do_not_release" -eq 1 ]; then
+   echo "Nothing to release"
+   exit 0
+else
+    # build and publish llama_cloud_services
```
adrianlyjak: some copy pasta here

.github/workflows/publish_changesets.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v3
+        with:
+          version: 10
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
adrianlyjak, line 31: seems rather old

scripts/test_new_version.py
```diff
@@ -0,0 +1,29 @@
+import toml  # type: ignore [import]
```
adrianlyjak: this test/validation doesn't seem very useful

adrianlyjak (approved):

PR #208, opened 2025-11-17
edits

adrianlyjak:


packages/llama-index-utils-workflow/package.json
```diff
@@ -1,14 +1,8 @@
 {
-  "name": "@llama-agents-workspace/llama-index-utils-workflow",
+  "name": "llama-index-utils-workflow",
   "version": "0.5.0",
-  "private": "true",
+  "private": "false",
```
AstraBert, line 4: I don't know if public or private makes a big difference here, since we don't publish this as a JS package nevertheless, but the ratio behind setting it as private is because this is not a JS package and should not be interpreted as such by changesets

adrianlyjak: The release process is completely run by our script, so this doesn't get published.

Maybe silly, but I'm reading the package.json private value in `changesets.py` to infer whether the package should be published (Could instead just treat all folders with both pyproject.toml and package.json as ones that should be published)

AstraBert: Oh yeah, makes sense then, fine to keep it as public :))

PR #209, opened 2025-11-17
Switch to uv build backend from hatchling

adrianlyjak:
Updates to uv build backend, is general release now, and seems more maintained than hatchling. We don't use anything fancy, so this should be perfectly safe.

Additionally updates CONTRIBUTING.md to mention changesets


CONTRIBUTING.md
```diff
@@ -17,23 +17,6 @@ Generally, we're looking for:
 
 Ideas beyond the above are welcome! This list is not exhaustive.
 
-## Code Structure
```
logan-markewich, line 20: lol

logan-markewich (approved):
lgtm

CONTRIBUTING.md
adrianlyjak: heh, my opinion is that if you are committing changes, you should be able to navigate the files. Also, this has gotten rather out of date

PR #224, opened 2025-11-20
chore: move representation utils to workflows core

AstraBert:
Move `_extract_workflow_structure` out of `llama-index-utils-workflow`, unify it with the one in workflows.server and move it to its own module

### DIFF
--- a/packages/llama-index-workflows/tests/test_representation_utils.py
+++ b/packages/llama-index-workflows/tests/test_representation_utils.py
@@ -75,24 +75,17 @@ def ground_truth_repr() -> DrawWorkflowGraph:
     )
 
 
-def test_extract_workflow_structure() -> None:
+def test_extract_workflow_structure(ground_truth_repr: DrawWorkflowGraph) -> None:
     wf = DummyWorkflow()
     graph = extract_workflow_structure(workflow=wf)
     assert isinstance(graph, DrawWorkflowGraph)
-    assert sorted(["end_step", "middle_step", "start_step"]) == sorted(
-        [node.id for node in graph.nodes if node.node_type == "step"]
-    )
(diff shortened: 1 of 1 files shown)


packages/llama-index-utils-workflow/src/llama_index/utils/workflow/__init__.py
```diff
@@ -15,194 +14,25 @@
...
 )
 from workflows.handler import WorkflowHandler
+from workflows.representation_utils import (
+    DrawWorkflowEdge,
+    DrawWorkflowGraph,
+    DrawWorkflowNode,
+    _truncate_label,
+)
+from workflows.representation_utils import (
+    extract_workflow_structure as _extract_workflow_structure,
```
logan-markewich, line 30: oh wow, the utils has the same output structure?

AstraBert: Yes, it just has some better typing but the same output

packages/llama-index-workflows/tests/test_representation_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    DrawWorkflowGraph,
+    DrawWorkflowNode,
+    extract_workflow_structure,
+)
+
+from .conftest import DummyWorkflow, LastEvent, OneTestEvent  # type: ignore[import]
+
+
+@pytest.fixture()
+def ground_truth_repr() -> DrawWorkflowGraph:
```
adrianlyjak, line 14: this appears to be unreferenced? Seems like a good test though, just dump a workflow's representation and compare the whole thing

AstraBert: Oh yeah, I used it as a reference to build the test but totally forgot to add it in the test

PR #246, opened 2025-12-16
New Turkish version README.md

Yigtwxx:
Translated the complete project documentation into Turkish. The content was localized while maintaining technical terminology and formatting.


logan-markewich:
I appreciate the contribution @Yigtwxx but I think we will keep the top-level README in english. Users can always throw this into a translator if needed 🙏🏻

PR #251, opened 2025-12-20
wip child workflows

logan-markewich:


packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -259,9 +259,249 @@ def add_step(cls, func: StepFunction) -> None:
...
+        # Get all steps from child workflow
+        child_steps = child._get_steps()
+
+        # Identify the boundary steps (those that accept StartEvent or return StopEvent)
+        # Note: A step can be BOTH a start and stop step if it directly accepts
+        # StartEvent and returns StopEvent
+        categorized_steps = []  # List of (step_name, step_func, is_start, is_stop)
+
+        for step_name, step_func in child_steps.items():
+            if step_name == "_done":
```
adrianlyjak, line 330: I think I got rid of this, but must have missed some spots (or maybe in claude training data?)

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -259,9 +259,249 @@ def add_step(cls, func: StepFunction) -> None:
...
+
+        The child workflow can still be run standalone, and when included, all its
+        internal steps execute normally while the parent's Context flows through.
+        """
+        # Use workflow class name as default namespace
+        if namespace is None:
+            namespace = workflow_class.__name__
+
+        # Create a temporary instance to introspect the child workflow
+        # TODO: how can we skip this instantiation?
```
adrianlyjak, line 309: seems like this should be doable to add to the meta class. It already exposes the steps function as a method

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -259,9 +259,249 @@ def add_step(cls, func: StepFunction) -> None:
 
         cls._step_functions[func.__name__] = func
 
+    def include_workflow(
```
adrianlyjak, line 262: This looks like an instance method. It'd be nice to make this declarative only. Easier to reason about and less edges (the add step thing is a pain, especially as we move towards distributed. Too dynamic). I'd want to be able to introspect the workflow class including child steps, rather than waiting for instantiation.

Maybe we can do this via a special step return value? Or via some class attribute?

```
class MyWorkflow(Workflow):
   child: ChildWorkflow # maybe this works and is enough? maybe weird to leave it undefined... could maybe auto instantiate if it doesn't have init params?

   # maybe have some special pattern for the constructor method? Could accept the injected deps too?
   @child_workflow
   def my_child(self, ctx: Context) -> ChildWorkflow:
        return ChildWOrkflow()

logan-markewich: Hmm, we can probably make this work

PR #261, opened 2026-01-06
Add idle workflow tracking via internal event

adrianlyjak:
## Add idle workflow tracking

Implements tracking for when workflows become idle (waiting on external input via `ctx.wait_for_event()`).

### Changes

- **New `WorkflowIdleEvent`**: Emitted when a workflow transitions to idle state
- **Idle detection**: Added `_check_idle_state()` helper to determine if all steps have no pending work but have active waiters
- **Server tracking**: `_WorkflowHandler` and `PersistentHandler` now track `idle_since` timestamp
- **Resumption detection**: `idle_since` is cleared when workflow resumes (receives `StepStateChanged(RUNNING)`)

### Definition of "Idle"

A workflow is idle when:
1. The workflow is running (not completed/failed/cancelled)
2. All steps have empty queues
3. All steps have no in-progress workers
4. At least one step has an active waiter

### Future Work

This enables future implementation of:
- Idle timeout configuration
- Releasing idle workflows from memory while preserving state
- Resume-on-event for released workflows

### DIFF
--- a/.changeset/open-boats-kick.md
+++ b/.changeset/open-boats-kick.md
@@ -1,5 +1,5 @@
 ---
-"llama-index-workflows": minor
+"llama-index-workflows": patch
 ---
 
 Track when workflows are idle (waiting on external input)


packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -542,6 +562,14 @@ def _process_step_result_tick(
                 event, tick.step_name, worker_state, now_seconds
             )
             commands.extend(subcommands)
+
+    # Check for idle transition at end of processing
+    was_idle = _check_idle_state(init)
```
logan-markewich, line 567: does init ever change? We are checking this on every tick

logan-markewich (approved):
This makes sense to me

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
adrianlyjak: Yeah, that's the initial state of the reducer function call, so its effectively the "last" state

Issue #222, opened 2025-11-19
Include timeout exception in StopEvent.result for streaming workflows


ttozser:
I have a workflow that streams events to the end user.
If the workflow times out, I receive an empty StopEvent without an indication of what the problem was.
I would like to receive the exception in the result field of the StopEvent so I can communicate the problem to the user.

-------------

For example:
```
exception = WorkflowTimeoutError(f"Operation timed out after {tick.timeout} seconds. {steps_info}")
return state, [
        CommandPublishEvent(event=StopEvent(result=exception)),
        CommandHalt(exception=exception),
    ]
```

I found these places with similar StopEvent. I think the exception should be provided in these places too.

https://github.com/AODocs-Dev/workflows-py/blob/300fd0541603d2496e6080bf9499205a0058df1c/packages/llama-index-workflows/src/workflows/runtime/control_loop.py#L438
https://github.com/AODocs-Dev/workflows-py/blob/300fd0541603d2496e6080bf9499205a0058df1c/packages/llama-index-workflows/src/workflows/runtime/control_loop.py#L653
https://github.com/AODocs-Dev/workflows-py/blob/300fd0541603d2496e6080bf9499205a0058df1c/packages/llama-index-workflows/src/workflows/runtime/control_loop.py#L681

adrianlyjak:
@ttozser I feel the pain here and have been planning on us implementing something to this effect. Agree that adding some information to these generic stop events would be a good direction to make it more informative. I have some reservations about putting it in the result field? There's some odd legacy semantics around StopEvents vs custom stop events and whether they "unwrap" the result in the workflow return value for compatibility reasons.

One possible idea to avoid building on that would be to rather have specific subclasses of StopEvent that instead have types and specific fields related to the failure mode of the workflow. For example a `WorkflowTimedOutEvent` event with the duration of the time out, or `WorkflowHaltedEvent`, and `WorkflowFailedEvent` with a serializable representation of the exception. Would love your feedback!

ttozser:
Having dedicated StopEvent subclasses is a good idea. 
It solves my issue, and it has a lower chance of creating compatibility issues.

PR #264, opened 2026-01-06
feat: add dedicated StopEvent subclasses for workflow termination scenarios

adrianlyjak:
Add WorkflowTerminationEvent base class and specific subclasses to provide
detailed information when workflows end abnormally:

- WorkflowTimedOutEvent: Published when workflow exceeds timeout, includes
  timeout duration and list of active steps
- WorkflowCancelledEvent: Published when workflow is cancelled by user
- WorkflowFailedEvent: Published when a step fails permanently, includes
  step name, exception type, and exception message

These events replace the empty StopEvent() instances that were previously
published, allowing consumers to understand why a workflow ended and take
appropriate action.

Fixes https://github.com/run-llama/workflows-py/issues/222

### DIFF
--- a/docs/api_docs/docs/api_reference/events.md
+++ b/docs/api_docs/docs/api_reference/events.md
@@ -6,3 +6,6 @@
         - HumanResponseEvent
         - StartEvent
         - StopEvent
+        - WorkflowTimedOutEvent
+        - WorkflowCancelledEvent
+        - WorkflowFailedEvent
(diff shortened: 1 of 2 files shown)


packages/llama-index-workflows/src/workflows/events.py
```diff
@@ -209,6 +209,71 @@ def __str__(self) -> str:
...
+        ```python
+        async for event in handler.stream_events():
+            if isinstance(event, WorkflowFailedEvent):
+                print(f"Step '{event.step_name}' failed: {event.exception_message}")
+        ```
+    """
+
+    step_name: str
+    exception_type: str
+    exception_message: str
```
logan-markewich, line 279: is the trace ever logged when this happens? Or, I guess the trace happens when `await handler` is called?

adrianlyjak: There'd be some sort of a stack trace. I could grab some more info, just want to keep it serializable

PR #266, opened 2026-01-07
drop and restore idle workflows in workflow server

### LINKED ISSUE #191: Support for Persistent Workflow Suspension and Resumption (for HITL event pauses)
In real-world use cases, human-in-the-loop responses can often take hours or even days to arrive. Currently, the workflow server is required to keep workflow state in memory while waiting for such responses, which is inefficient and limits scalability.

We propose introducing the ability to pause (persist) a workflow’s state to the workflow store when awaiting human input, rather than holding it in memory. Once the human response event is received, the workflow should be able to resume seamlessly from the persisted state.

This enhancement would improve reliability, scalability, and fault tolerance for long-running workflows involving asynchronous human participation.

adrianlyjak:
Updates workflow server with functionality to drop and restore idle workflow handlers that are waiting on external input.

Adds an additional internal event to detect failed messages, such that a workflow can be returned to idle for an edge case where a non-handled event could awaken the handler without notifying that it was still actually idle

Fixes #191

### DIFF
--- a/packages/llama-index-workflows/tests/server/test_idle_release_live_http.py
+++ b/packages/llama-index-workflows/tests/server/test_idle_release_live_http.py
@@ -4,22 +4,19 @@
 
 from __future__ import annotations
 
-import asyncio
-import socket
-from contextlib import asynccontextmanager, closing
 from datetime import timedelta
-from typing import AsyncGenerator
 
-import httpx
 import pytest
-import uvicorn
 from workflows import Context, Workflow, step
 from workflows.client.client import WorkflowClient
 from workflows.events import Event, StartEvent, StopEvent, WorkflowIdleEvent
 from workflows.server import WorkflowServer
 from workflows.server.memory_workflow_store import MemoryWorkflowStore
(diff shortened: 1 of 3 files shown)


packages/llama-index-workflows/src/workflows/server/server.py
```diff
@@ -931,14 +944,19 @@ async def _stream_events(self, request: Request) -> StreamingResponse:
...
-            persisted = await self._workflow_store.query(
-                HandlerQuery(handler_id_in=[handler_id])
-            )
-            if persisted:
-                status = persisted[0].status
-                if status in {"completed", "failed", "cancelled"}:
-                    raise HTTPException(detail="Handler is completed", status_code=204)
-            raise HTTPException(detail="Handler not found", status_code=404)
+            # Try to reload from persistence (for released idle workflows)
+            handler = await self._try_reload_handler(handler_id)
```
adrianlyjak: The double persistence query here is not ideal

PR #269, opened 2026-01-08
Add resource nodes to workflow graph export

adrianlyjak:
Add support for visualizing resource dependencies in workflow graphs:

- Enhanced _Resource class with source location metadata (file, line, docstring, unique hash for deduplication)
- Added ResourceDefinition.type_annotation to capture the type from Annotated[T, Resource(...)]
- Added WorkflowGraphResourceNode protocol model with full metadata
- Added DrawWorkflowResourceNode class for intermediate representation
- Updated DrawWorkflowEdge to support edge labels (variable names)
- Modified extract_workflow_structure to extract and deduplicate resource nodes, connecting them to steps with labeled edges
- Updated Pyvis and Mermaid renderers to display resource nodes (hexagon shape, plum color) with metadata tooltips

Mermaid

image

Vis html

image

Additionally:
- Removed excessive duplication of workflow graph dataclasses with pydantic models
- Split nodes into a discriminated union of subtypes for better attributing fields to a node type
- added doc strings as descriptions to workflows and nodes where applicable
- add schema information to event nodes
- added a utility for condensing the graph, removing an unwanted type (so e.g. you can now remove the new resource type, or, if you want you could convert the event node into just an edge

### DIFF
--- a/packages/llama-index-workflows/src/workflows/protocol/__init__.py
+++ b/packages/llama-index-workflows/src/workflows/protocol/__init__.py
@@ -142,10 +142,6 @@ class WorkflowResourceNode(WorkflowNodeBase):
         default=None,
         description="Documentation string extracted from the getter function",
     )
-    unique_hash: str | None = Field(
-        default=None,
-        description="Hash used for deduplication when the same resource is used in multiple steps",
-    )
 
 
 class WorkflowGenericNode(WorkflowNodeBase):
--- a/packages/llama-index-workflows/src/workflows/representation_utils.py
+++ b/packages/llama-index-workflows/src/workflows/representation_utils.py
@@ -89,7 +89,6 @@ def _create_resource_node(resource_def: ResourceDefinition) -> WorkflowResourceN
         source_file=source_file,
         source_line=source_line,
         description=resource_description,
-        unique_hash=unique_hash,
     )
 
 
--- a/packages/llama-index-workflows/tests/test_representation_utils.py
+++ b/packages/llama-index-workflows/tests/test_representation_utils.py
@@ -191,7 +191,6 @@ def test_extract_workflow_structure_with_resources() -> None:
     assert "Factory function" in resource_node.description
     assert resource_node.source_file is not None
     assert resource_node.source_line is not None
-    assert resource_node.unique_hash is not None
 
 
 def test_resource_node_edges_have_variable_names() -> None:
@@ -298,7 +297,6 @@ def test_resource_node_serialization() -> None:
         source_file="/path/to/file.py",
         source_line=42,
         description="Test docstring",
-        unique_hash="abc123",
     )
 
     assert resource_node.id == "resource_abc123"
@@ -309,7 +307,6 @@ def test_resource_node_serialization() -> None:
     assert resource_node.source_file == "/path/to/file.py"
     assert resource_node.source_line == 42
     assert resource_node.description == "Test docstring"
-    assert resource_node.unique_hash == "abc123"
 
     # Test serialization
     data = resource_node.model_dump()
@@ -431,7 +428,6 @@ def test_resource_node_serialization_roundtrip() -> None:
         source_file="/path/to/source.py",
         source_line=100,
         description="Resource docstring",
-        unique_hash="abc123",
     )
 
     data = node.model_dump()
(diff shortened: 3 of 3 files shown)


logan-markewich (approved):

examples/visualization/resource_nodes_example.py
```diff
@@ -0,0 +1,260 @@
...
+def get_database_client() -> DatabaseClient:
+    """Factory function to create a database client.
+
+    This function creates a PostgreSQL database client configured
+    for the application's data storage needs.
+    """
+    return DatabaseClient(connection_string="postgres://localhost/myapp")
+
+
+def get_cache_client() -> CacheClient:
```
AstraBert, line 80: Small nit: I think when we first designed resources we agreed that it would be a good idea to use resource functions with `*args, **kwargs` for backward compatibility in case we ever wanted to do something with the function within the resource itself (non-blocking)

packages/llama-index-utils-workflow/tests/test_drawing.py
```diff
@@ -309,3 +309,162 @@ async def test_draw_most_recent_execution_mermaid(workflow: Workflow) -> None:
         edge_lines = [line for line in lines if " --> " in line]
         assert len(node_lines) > 0
         assert len(edge_lines) > 0
+
+
+# --- Resource node rendering tests ---
+
+
+@pytest.mark.asyncio
+async def test_mermaid_resource_nodes_rendered(
```
AstraBert: Why async here? And in the other tests, they don't seem to be using async functions

packages/llama-index-workflows/src/workflows/protocol/__init__.py
```diff
@@ -62,16 +62,30 @@ class WorkflowGraphResponse(BaseModel):
...
+    Represents steps, events, external nodes, or resources.
+    Resource-specific fields are optional and only populated for resource nodes.
+    """
+
     id: str
     label: str
-    node_type: str
-    title: str | None
-    event_type: str | None
+    node_type: str  # 'step', 'event', 'external', 'resource'
```
AstraBert: Could we make this a literal? I know it was a string also before, but maybe it would be more self-explanatory if it was a literal

packages/llama-index-workflows/src/workflows/protocol/__init__.py
```diff
@@ -62,16 +62,30 @@ class WorkflowGraphResponse(BaseModel):
...
+    title: str | None = None
+    event_type: str | None = None
+    # Resource-specific fields (only populated when node_type == "resource")
+    type_name: str | None = None  # The type annotation, e.g., "AsyncLlamaCloud"
+    getter_name: str | None = None  # The factory function name
+    source_file: str | None = None  # Path to the source file containing the getter
+    source_line: int | None = None  # Line number where the getter is defined
+    docstring: str | None = None  # Documentation string of the getter function
+    unique_hash: str | None = None  # Unique identifier for deduplication
```
AstraBert: Could it be a good idea to put all of this in a `ResourceNodeMetadata` model and create a `resource_metadata: ResourceNodeMetadata | None = None` field here?
And we could add a model validator that checks that, when a node's type is `resource`, the `resource_metadata` field is not null.

Something like:

```python
@model_validator(mode="after")
def validate_resource(self) -> Self:
    if self.node_type == "resource" and self.resource_metadata is None:
        raise ValueError("resource_metadata cannot be None if the node is a resource")
```

packages/llama-index-workflows/src/workflows/representation_utils.py
```diff
@@ -21,15 +24,26 @@
...
-    )
+    # Event-specific field
+    event_type: Optional[type] = None
+    # Resource-specific fields
+    type_name: Optional[str] = None  # e.g., "AsyncLlamaCloud"
+    getter_name: Optional[str] = None  # e.g., "get_llama_cloud_client"
+    source_file: Optional[str] = None
+    source_line: Optional[int] = None
+    docstring: Optional[str] = None
+    unique_hash: Optional[str] = None
```
AstraBert: Same comment as for the BaseModel, except that the validation logic here should be placed in `__post_init__`

AstraBert (commented):
Overall agree with this idea, left some minor comments for things I would like

adrianlyjak:
> Latest changes are lgtm!

Iterating further, those seemed a bit awkward. Down a rabbit hole now 🐰 🕳️ .

I think I'm going to remove the dataclass/pydantic duplication

examples/visualization/resource_nodes_example.py
adrianlyjak: My thoughts are that we should instead start to support recursive resources (and "resource configs"), similar to fastapi dependencies

PR #271, opened 2026-01-09
feat: first stab at configurable resource

AstraBert:
Add Pydantic-based resource that is configurable from a JSON file

### DIFF
--- a/packages/llama-index-workflows/src/workflows/resource.py
+++ b/packages/llama-index-workflows/src/workflows/resource.py
@@ -75,19 +75,21 @@ class _Resource(Generic[T]):
 def _get_resource_config_data(
     config_file: str,
     path_selector: str | None,
-) -> dict[str, Any] | None:
+) -> dict[str, Any]:
     with open(config_file, "r") as f:
         data = json.load(f)
     if path_selector is not None:
         keys = path_selector.split(".")
-        val: Optional[dict[str, Any]] = data  # noqa: UP045
+        val: dict[str, Any] = data
+        cumulative_path = ""
         for key in keys:
-            if isinstance(val, dict):
-                val = cast(Optional[dict[str, Any]], val.get(key))  # noqa: UP045
-                if val is None:
-                    return None
-            else:
-                return None
+            cumulative_path += key + "."
+            got = cast(Optional[dict[str, Any]], val.get(key))  # noqa: UP045
+            if not isinstance(got, dict):
+                raise ValueError(
+                    f"Expected dictionary for configuration from {config_file} at path {cumulative_path.strip('.')}, got: {type(got)}"
+                )
+            val = got
         return val
     return data
 
@@ -121,20 +123,16 @@ class _ResourceConfig(Generic[B]):
 
     # make async for compatibility with _Resource
     def call(self) -> B:
-        if (
-            sel_data := _get_resource_config_data(
-                config_file=self.config_file, path_selector=self.path_selector
-            )
-        ) is not None:
-            # let validation error bubble up
-            if self.cls_factory is not None:
-                return self.cls_factory.model_validate(sel_data)
+        sel_data = _get_resource_config_data(
+            config_file=self.config_file, path_selector=self.path_selector
+        )
+        # let validation error bubble up
+        if self.cls_factory is not None:
+            return self.cls_factory.model_validate(sel_data)
+        else:
             raise ValueError(
                 "Class factory should be set to a BaseModel subclass before calling"
             )
-        raise ValueError(
-            f"Invalid path selector for {self.config_file}: {self.path_selector}"
-        )
 
 
 def ResourceConfig(
--- a/packages/llama-index-workflows/tests/test_resources.py
+++ b/packages/llama-index-workflows/tests/test_resources.py
@@ -161,6 +161,45 @@ def test_resource_config_path_selector(
     assert value.dirs == ["hello/"]
 
 
+def test_resource_config_path_selector_error(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.chdir(tmp_path)
+    data = {
+        "core": {"fs": {"files": ["hello.py"], "dirs": ["hello/"]}},
+    }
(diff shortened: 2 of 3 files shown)


packages/llama-index-workflows/src/workflows/resource.py
```diff
@@ -34,15 +42,133 @@ def __init__(self, factory: Callable[..., T | Awaitable[T]], cache: bool) -> Non
...
+    @property
+    def name(self) -> str:
+        if self.path_selector is not None:
+            return self.config_file + "." + self.path_selector
+        return self.config_file
+
+    def _get_data(self, data: dict[str, Any]) -> Optional[dict[str, Any]]:  # noqa: UP045
+        if self.path_selector is not None:
+            keys = self.path_selector.split(".")
+            val: Optional[dict[str, Any]] = data  # noqa: UP045
```
adrianlyjak: What's the reason for Optional with the noqa? we should be using `| None` types rather than `Optional`.

packages/llama-index-workflows/src/workflows/resource.py
```diff
@@ -34,15 +42,133 @@ def __init__(self, factory: Callable[..., T | Awaitable[T]], cache: bool) -> Non
         self.name = getattr(factory, "__qualname__", type(factory).__name__)
         self.cache = cache
 
-    async def call(self) -> T:
+    def get_resource_configs(
+        self, resource_configs_cache: dict[str, dict[str, BaseModel]]
+    ) -> dict[str, BaseModel]:
+        params = inspect.signature(self._factory).parameters
```
adrianlyjak: I would try to optimize such that inspect isn't called on every get. Can this go in the constructor instead maybe? Or have a simple memoization?

packages/llama-index-workflows/src/workflows/resource.py
```diff
@@ -34,15 +42,133 @@ def __init__(self, factory: Callable[..., T | Awaitable[T]], cache: bool) -> Non
...
+    ) -> dict[str, BaseModel]:
+        params = inspect.signature(self._factory).parameters
+        resource_configs: dict[str, BaseModel] = {}
+        if len(params) > 0:
+            for param in params.values():
+                if get_origin(param.annotation) is Annotated:
+                    args = get_args(param.annotation)
+                    if len(args) == 2 and isinstance(args[1], _ResourceConfig):
+                        resource_config = args[1]
+                        if (
```
adrianlyjak: I think a simple lru cache function, and making caching always on would be simpler. I can't imagine a great use for disabling caching (updating configs or ginormous config objects leaking memory seem rare and not ideal fit for this concept)

packages/llama-index-workflows/tests/test_resources.py
```diff
@@ -93,6 +214,156 @@ def f1(
...
+        @step
+        def start_step(self, ev: StartEvent) -> SecondEvent:
+            print("Start step is done", flush=True)
+            return SecondEvent(msg="Hello")
+
+        @step
+        def f1(
+            self,
+            ev: SecondEvent,
+            file_operator: Annotated[FileOperator, Resource(get_file_operator)],
```
adrianlyjak, line 285: this quickly gets complicated, so probably a later thing, but I would like to move towards a model where 
- configs can be loaded from a step, not just from a resource
- resources can recursively depend on other resources

packages/llama-index-workflows/src/workflows/resource.py
AstraBert: mypy was complaining about type annotations not being explicit here, so I set `val: dict[str, Any] | None`, but then tests on python 3.9 failed because it does not support that notation. So I reverted to `Optional[dict[str, Any]]` and python 3.9 was happy, but ruff was saying I should use the `X | None` notation, so I added the noqa to turn off the ruff problem

packages/llama-index-workflows/src/workflows/resource.py
AstraBert: Oh yeah good point

packages/llama-index-workflows/tests/test_resources.py
AstraBert: Makes sense, and yeah, I would probably add it in a later PR

packages/llama-index-workflows/src/workflows/resource.py
AstraBert: Now the resource configs are computed once per instance :)

packages/llama-index-workflows/src/workflows/resource.py
AstraBert: Should be solved now

packages/llama-index-workflows/src/workflows/resource.py
adrianlyjak: Sad, I thought `from __future__ import annotations` fixed this, but maybe there's some special scenario here

packages/llama-index-workflows/src/workflows/resource.py
```diff
@@ -33,16 +42,120 @@ def __init__(self, factory: Callable[..., T | Awaitable[T]], cache: bool) -> Non
...
+    path_selector: str | None,
+) -> dict[str, Any] | None:
+    with open(config_file, "r") as f:
+        data = json.load(f)
+    if path_selector is not None:
+        keys = path_selector.split(".")
+        val: Optional[dict[str, Any]] = data  # noqa: UP045
+        for key in keys:
+            if isinstance(val, dict):
+                val = cast(Optional[dict[str, Any]], val.get(key))  # noqa: UP045
```
adrianlyjak: nit: this will lead to weird hard to interpret stack errors if the val is not actually a dict, (string, bool, etc)

adrianlyjak (approved):

packages/llama-index-workflows/src/workflows/resource.py
AstraBert: I will put an explicit error in this function rather than returning `None`, might be easier to trace back

AstraBert: Done now✅

PR #277, opened 2026-01-14
Separate server/client, and move to llama_agents namespace

adrianlyjak:


packages/llama-index-workflows-client/pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[build-system]
+requires = ["uv_build>=0.9.6,<0.10.0"]
+build-backend = "uv_build"
+
+[project]
+name = "llama-index-workflows-client"
+version = "0.1.0"
+description = "Add your description here"
```
AstraBert: Nit: package description would be nice

packages/llama-index-workflows-server/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[build-system]
+requires = ["uv_build>=0.9.6,<0.10.0"]
+build-backend = "uv_build"
+
+[project]
+name = "llama-index-workflows-server"
+version = "0.1.0"
+description = "Add your description here"
```
AstraBert: Description, as per before :))

AstraBert (approved):
Looks good to me!

examples/client/base/workflow_client.py
logan-markewich, line 1: do we want to change these examples yet or nah?

packages/llama-index-workflows/pyproject.toml
```diff
@@ -50,4 +53,5 @@ asyncio_mode = "auto"
 testpaths = ["tests"]
 
 [tool.uv.build-backend]
-module-name = "workflows"
+module-name = ["workflows", "llama_agents.workflows"]
```
logan-markewich, line 56: Is this all it takes to publish the two packages and support the two import paths? 🤯

examples/client/base/workflow_client.py
adrianlyjak: I figured yeah, its in the repo might as well keep it current

packages/llama-index-workflows/pyproject.toml
logan-markewich: Ohhhh I see, there's two folders under src

logan-markewich (approved):
Assuming this works, lets try it?

packages/llama-index-workflows/pyproject.toml
logan-markewich: And some sneaky stuff with extending exported paths

adrianlyjak: and a whole lot of typing ignores, the type checkers don't like it

PR #282, opened 2026-01-15
fix: rebuild_state_from_ticks clears in_progress before replaying

### LINKED ISSUE #281: ctx.to_dict() fails with 'Worker X not found in in_progress' when checkpointing resumed workflows
## Bug Description

When resuming a workflow from a checkpoint and then checkpointing again during the resumed run, `ctx.to_dict()` fails with:

```
ValueError: Worker 0 not found in in_progress
```

## Steps to Reproduce

1. Run a workflow with parallel workers (e.g., `num_workers > 1`)
2. Checkpoint while workers are in progress
3. Crash/stop the workflow
4. Resume from checkpoint
5. Let the resumed workflow continue
6. Try to checkpoint again (call `ctx.to_dict()`)
7. **Error:** "Worker X not found in in_progress"

## Root Cause

When resuming:
1. `BrokerState.from_serialized()` moves `in_progress` events to queue
2. `rewind_in_progress()` clears `in_progress` and assigns **new** worker IDs

When checkpointing the resumed workflow:
1. `ctx.to_dict()` calls `rebuild_state_from_ticks(init_state, tick_log)`
2. `init_state` still has original `in_progress` worker IDs from checkpoint
3. `tick_log` has new worker IDs assigned after rewind
4. **Mismatch causes the error**

## Expected Behavior

`ctx.to_dict()` should work correctly on resumed workflows, enabling crash recovery scenarios where you need to checkpoint multiple times.

## Environment

- workflows version: 2.x
- Python: 3.10+

OS-ramamurtisubramanian:
## Summary

Fixes `ctx.to_dict()` failing with "Worker X not found in in_progress" when checkpointing resumed workflows.

## Problem

When resuming a workflow from checkpoint:
1. `rewind_in_progress()` clears `in_progress` and assigns **new** worker IDs
2. New ticks reference these new worker IDs
3. But `rebuild_state_from_ticks()` replayed ticks against the **original** `init_state`
4. Worker ID mismatch caused the error

## Solution

Modified `rebuild_state_from_ticks()` to clear `in_progress` and move events back to the queue before replaying ticks. This mirrors what `rewind_in_progress()` does at runtime, ensuring the baseline state aligns with what the control loop uses.

## Changes

- `control_loop.py`: Updated `rebuild_state_from_ticks()` to clear `in_progress` before replaying
- `test_control_loop_transformations.py`: Added 2 test cases

## Test Plan

- [x] `test_rebuild_state_from_ticks_clears_in_progress` - verifies the bug fix
- [x] `test_rebuild_state_from_ticks_preserves_queue_order` - verifies retry metadata preserved
- [x] All 39 existing control loop transformation tests pass

Closes #281

### DIFF
--- a/packages/llama-index-workflows/tests/test_workflow.py
+++ b/packages/llama-index-workflows/tests/test_workflow.py
@@ -1076,7 +1076,7 @@ async def test_workflow_parallel_resume() -> None:
             return ParDone(id=ev.id)
 
         @step
-        async def step3(self, ev: ParDone, ctx: Context) -> StopEvent | None:
+        async def step3(self, ev: ParDone, ctx: Context) -> Optional[StopEvent]:  # noqa - python 3.9 struggles here with | None
             if ev.id == allowed_index:
                 allowed_done.set()
             if ctx.collect_events(ev, [ParDone] * 4) is None:


adrianlyjak:
@OS-ramamurtisubramanian thanks for the bug and PR! I just pushed a more end-to-end test, and am still seeing it fail with the same reported error. I'm also seeing a type error from the pre-commit checks. Wondering if there's somewhere better to add this modification, I'm not sure the replay_from_ticks is the best place--it would be nice if that reflected the true execution, so I'm wondering if there's somewhere else like the rewind in progress that might just need further clean up

adrianlyjak (approved):

PR #285, opened 2026-01-16
feat: Add internal workflow registry to facilitate workflow discovery…

adrianlyjak:
… for runtime plugins


adrianlyjak:
nah, this is the wrong direction

PR #287, opened 2026-01-19
docs: add Context import to typed state workflow example

mrbadri:
## Description

Added the missing `Context` import to the typed state example in the state management docs so the example works as expected when copied and run.

This is a documentation-only change and does not affect behavior or APIs.


adrianlyjak:
@mrbadri thanks again! I did an audit for doc import issues, and already got this fix in ahead of you 😂

PR #288, opened 2026-01-19
docs: fix streamed response handling in streaming events example

mrbadri:
## Description
Fixes the streaming events example to return the full accumulated LLM response instead of only the last streamed chunk. This makes the final workflow output consistent with the streamed progress shown during execution.


adrianlyjak:
@mrbadri good catch. Thank you!

mrbadri:
> @mrbadri good catch. Thank you!

You’re welcome! Glad it helped.

PR #291, opened 2026-01-20
Draw all possible nested workflows

hippopond:
Previously, when a user has a nested workflow (childworkflow is instantiated in one of the Steps in a Parent Workflow):   e.g.
```
####################################################################
class ChildWorkflowA(Workflow):
    @step()
    async def child_start(self, ev: StartEvent) -> StopEvent:
        return StopEvent(result="Child processed")


class ParentWorkflow(Workflow):
    @step()
    async def parent_start(self, ev: StartEvent) -> Event:
        # Instantiate the nested workflow. The drawing function will inspect this.
        child_wf = ChildWorkflowA()
        # The actual run logic is not important for drawing.
        _ = await child_wf.run(input="dummy")
        return Event(result="some result")

    @step()
    async def parent_end(self, ev: Event) -> StopEvent:
        return StopEvent(result="Final Result")
####################################################################
```
The utility function: draw_all_possible_flows_mermaid
draws only the parent workflow with no indication of the child.   

This PR introduces a new function (with the corresponding test case) :  draw_all_possible_flows_nested_mermaid
that draws the second level nested Workflow.  

NOTE: 
- 1 Nested Level. (currently not recursive) 
- If the same ChildWorkflow is used in a separate Parent Step, it is treated separately (I can make it the same) 

Here is a side by side of what the difference is for the 2 utility functions (existing and new addition): 

image


hippopond:
It passes precommit.  

Let me add a changeset

hippopond:
I am cancelling this PR and making a new PR.   (previously my local environment was not right).   Will try again

PR #292, opened 2026-01-22
Draw nested workflow mermaid

hippopond:
Previously, when a user has a nested workflow (childworkflow is instantiated in one of the Steps in a Parent Workflow):   e.g.
```
####################################################################
class ChildWorkflowA(Workflow):
    @step()
    async def child_start(self, ev: StartEvent) -> StopEvent:
        return StopEvent(result="Child processed")


class ParentWorkflow(Workflow):
    @step()
    async def parent_start(self, ev: StartEvent) -> Event:
        # Instantiate the nested workflow. The drawing function will inspect this.
        child_wf = ChildWorkflowA()
        # The actual run logic is not important for drawing.
        _ = await child_wf.run(input="dummy")
        return Event(result="some result")

    @step()
    async def parent_end(self, ev: Event) -> StopEvent:
        return StopEvent(result="Final Result")
####################################################################
```
The utility function: draw_all_possible_flows_mermaid
draws only the parent workflow with no indication of the child.   

This PR introduces a new function (with the corresponding test case) :  draw_all_possible_flows_nested_mermaid
that draws the second level nested Workflow.  

NOTE: 
- 1 Nested Level. (currently not recursive) 
- If the same ChildWorkflow is used in a separate Parent Step, it is treated separately (I can make it the same) 

Here is a side by side of what the difference in output is for the 2 utility functions (existing and new addition): 

side_by_side

### DIFF
--- a/packages/llama-index-utils-workflow/src/llama_index/utils/workflow/__init__.py
+++ b/packages/llama-index-utils-workflow/src/llama_index/utils/workflow/__init__.py
@@ -48,6 +48,8 @@ def _get_node_color(node: WorkflowGraphNode) -> str:
         return "#ADD8E6"  # Light blue for steps
     elif node.node_type == "external":
         return "#BEDAE4"  # Light blue-gray for external
+    elif node.node_type == "child_connector":
+        return "#E0E0E0"  # Light gray for child workflow connectors
     elif node.node_type == "resource":
         return "#DDA0DD"  # Plum/light purple for resources
     elif node.node_type == "resource_config":
@@ -82,6 +84,8 @@ def _get_node_shape(node: WorkflowGraphNode) -> str:
     """Determine shape for a node based on its type."""
     if node.node_type in ("step", "external"):
         return "box"
+    elif node.node_type == "child_connector":
+        return "ellipse"
     elif node.node_type == "event":
         return "ellipse"
     elif node.node_type == "resource":
@@ -187,6 +191,8 @@ def _get_mermaid_css_class(node: WorkflowGraphNode) -> str:
         return "stepStyle"
     elif node.node_type == "external":
         return "externalStyle"
+    elif node.node_type == "child_connector":
+        return "childConnectorStyle"
     elif node.node_type == "resource":
         return "resourceStyle"
     elif node.node_type == "resource_config":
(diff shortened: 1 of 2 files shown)


adrianlyjak:
Neat! The AST inspection code is actually surprisingly clean. 

Just heads up, we do plan on adding more official child workflows in the future, with some better communication patterns between parent/child. Once proper child workflows are implemented, we'd likely replace this source inspection functionality with the official child workflows (they'll be in the type signatures somehow).

packages/llama-index-utils-workflow/src/llama_index/utils/workflow/__init__.py
```diff
@@ -722,3 +726,118 @@ def draw_most_recent_execution_mermaid(
...
+        return []
+    return workflow_classes
+
+
+def draw_all_possible_flows_nested_mermaid(
+    workflow: Workflow,
+    filename: str = "workflow_nested_flows.mermaid",
+    max_label_length: int | None = None,
+) -> str:
+    parent_graph = _get_workflow_representation(workflow)
```
adrianlyjak: Ideally we'd make this process of defining the child workflows a bit more shareable. Could we make a new function that runs `_get_workflow_representation`, and then does the child workflow lookup and maybe extends the base data model with the new nodes? They potentially might be able to just drop in after adding the edge from `step -> workflow`

I'm also not sure I see the need to make this a separate function `draw_all_possible_flows_nested_mermaid`. Seems like reasonable default behavior to add to `draw_all_possible_flows_mermaid`, perhaps with a function flag to disable or enable it, (`include_child_workflows: bool = True`).

hippopond: 1) That makes sense.   If I understand that correctly, refactoring it that way would lend itself for recursion (in the future) and for other formats (pyvis, viz etc).   

2) It did cross my mind on the optional argument for include_child_workflows.   Glad you clarified that.  

Let me iterate on the above 2 points and turnaround.  


Ack: on the future changes on childworkflows

hippopond:
Refactored: draw_all_possible_flows_mermaid 
-----------------------------------------------
Extracted graph construction logic into a reusable _get_nested_workflow_representation helper.


Helper: _get_nested_workflow_representation
----------------------------------------------
This helper utilizes a closure (called: _merge_subgraph_into_parent) to handle complex graph stitching and ID prefixing, ensuring a clean and encapsulated state management during the merge.   


Helper: _get_workflow_classes_from_step
------------------------------------------
Additionally, implemented a robust  utility that uses AST parsing and type inspection to detect nested sub-workflows regardless of their class names.

adrianlyjak:
Updating with support for pyvis, and converting the "calls" edge to a node so that it lays out better in pyvis

image

image

hippopond:
Thank you !

PR #293, opened 2026-01-22
add llama_index integration tests

adrianlyjak:
Copied adapted llama index core workflows tests to ensure that they run against any  core workflows updates.

Also updated ANN checks to run, to enforce better typing in tests. Adds a bunch of check ignores for existing issues that were being ignored

### DIFF
--- a/packages/llama-index-integration-tests/src/llama_index_integration_tests/helpers.py
+++ b/packages/llama-index-integration-tests/src/llama_index_integration_tests/helpers.py
@@ -1,5 +1,7 @@
 """Test helpers for llama-index workflow integration tests."""
 
+from __future__ import annotations
+
 from typing import Callable, List
 
 from llama_index.core.base.llms.types import (
--- a/packages/llama-index-integration-tests/tests/conftest.py
+++ b/packages/llama-index-integration-tests/tests/conftest.py
@@ -1,5 +1,7 @@
 """Shared fixtures for llama-index workflow integration tests."""
 
+from __future__ import annotations
+
 from typing import Any, Callable, List, Protocol, Union
 
 import pytest


packages/llama-index-integration-tests/src/llama_index_integration_tests/helpers.py
```diff
@@ -0,0 +1,48 @@
...
+        additional_kwargs={
+            "tool_calls": [
+                ToolSelection(
+                    tool_id=tool_id,
+                    tool_name=tool_name,
+                    tool_kwargs=tool_kwargs or {},
+                )
+            ]
+        },
+    )
```
AstraBert, line 45: I think this should be expressed with `ToolCallBlock` now, something along the lines of:


```suggestion
    return ChatMessage(
        role=MessageRole.ASSISTANT,
        content=[TextBlock(text=content), ToolCallBlock(tool_call_id=tool_id, tool_name=tool_name, tool_kwargs=tool_kwargs or {})],
    )
```

packages/llama-index-integration-tests/tests/test_context_store.py
```diff
@@ -0,0 +1,177 @@
...
+    create_workflow: WorkflowFactory,
+) -> None:
+    """Test that initial_state is accessible via ctx.store in tools."""
+    received_value = None
+
+    async def check_state(ctx: Context) -> str:
+        nonlocal received_value
+        state = await ctx.store.get("state")
+        received_value = state.get("initial_key")
+        return f"Got: {received_value}"
```
AstraBert, line 26: This makes me think that we should probably be using the context state instead of the store directly

adrianlyjak:
> Looks good, we should add CI for the new tests

It should be running, we’re kind of just installing all the packages and running pytest against everything simultaneously. There’s some not great things about dependency isolation with that, but it runs fast and makes ci super simple

AstraBert:
> It should be running, we’re kind of just installing all the packages and running pytest against everything simultaneously. There’s some not great things about dependency isolation with that, but it runs fast and makes ci super simple

Mhhh, I don't see the tests being run for `llama-index-integration-tests` in the CI run here, not sure why

Edit: I think you should probably add the package in the test workflow :)

adrianlyjak:
> Mhhh, I don't see the tests being run for llama-index-integration-tests in the CI run here, not sure why

😅 you are correct, I completely forgot how I set this up 🤦

packages/llama-index-integration-tests/tests/test_context_store.py
adrianlyjak: No, the store is currently the recommended way to access. Its a little not great, I think this comes more from `AgentWorkflow`, which has the special `initial_state` constructor that initializes the "state" key _of_ the state store, so its more or less `state.state` 😵‍💫 

https://github.com/run-llama/llama_index/blob/e5c8d29f880abe64cc9d58382bc57c2e43ccd4fc/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py#L285

PR #294, opened 2026-01-22
Add resource config node support to workflow representation

adrianlyjak:
## Summary
This PR adds support for representing resource configuration nodes in workflow graphs. Resource configs are now extracted from resource factory functions and visualized as dedicated nodes in the workflow representation, with edges connecting resources to their configurations.

## Key Changes

- **New `WorkflowResourceConfigNode` type**: Added a new node type to represent configuration objects loaded from JSON files, including:
  - Configuration file path and optional path selector for nested values
  - Pydantic JSON schema for the config type (if applicable)
  - Actual config values read from files
  
- **Resource config extraction**: Implemented `_extract_resource_configs()` to parse resource factory functions and identify `ResourceConfig` annotations, extracting:
  - Parameter names and type information
  - Config file paths and path selectors
  - JSON schema from Pydantic BaseModel types
  - Actual config values from files

- **Graph building enhancements**:
  - Resource config nodes are automatically added to workflow graphs
  - Deduplication by (config_file, path_selector) tuple prevents duplicate nodes
  - Edges from resources to configs are labeled with the parameter name
  
- **Comprehensive test coverage**: Added 10 new tests covering:
  - Basic resource config node extraction
  - Path selectors for nested configs
  - Config deduplication across multiple resources
  - Multiple different configs in one workflow
  - Node serialization/deserialization
  - Graph filtering with resource config nodes

## Implementation Details

- Config values are read from JSON files with graceful error handling (returns None if file unreadable)
- Resource config nodes use SHA256 hashing of (config_file, path_selector) for unique, deterministic IDs
- Pydantic `model_json_schema()` is used to extract JSON schemas from BaseModel config types
- The implementation maintains backward compatibility with existing workflow representations

---

<a href="https://app.devin.ai/review/run-llama/workflows-py/pull/294">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/llama-index-workflows/src/workflows/representation/build.py
+++ b/packages/llama-index-workflows/src/workflows/representation/build.py
@@ -100,14 +100,10 @@ def _create_resource_config_node(
         model_cls: type[BaseModel] = type_annotation
         config_schema = model_cls.model_json_schema()
 
-    # Read config value using existing infrastructure (with error handling)
-    config_value: dict[str, Any] | None = None
-    try:
-        config_value = _get_resource_config_data(
-            resource_config.config_file, resource_config.path_selector
-        )
-    except (OSError, ValueError):
-        pass  # File not readable or path selector invalid
+    # Read config value using existing infrastructure
+    config_value = _get_resource_config_data(
+        resource_config.config_file, resource_config.path_selector
+    )
 
     return WorkflowResourceConfigNode(
         id=node_id,
@@ -137,14 +133,8 @@ def _create_resource_node(resource_def: ResourceDefinition) -> WorkflowResourceN
 
     if isinstance(resource, _Resource):
         factory = resource._factory
-        try:
-            source_file = inspect.getfile(factory)  # type: ignore[arg-type]
-        except (TypeError, OSError):
-            pass
-        try:
-            _, source_line = inspect.getsourcelines(factory)  # type: ignore[arg-type]
-        except (TypeError, OSError):
-            pass
+        source_file = inspect.getfile(factory)  # type: ignore[arg-type]
+        _, source_line = inspect.getsourcelines(factory)  # type: ignore[arg-type]
         resource_description = inspect.getdoc(factory)
 
     # Compute unique hash for deduplication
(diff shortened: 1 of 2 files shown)


packages/llama-index-workflows/src/workflows/representation/build.py
```diff
@@ -62,22 +76,57 @@ def _get_event_type_chain(cls: type) -> list[str]:
...
+        config_schema = model_cls.model_json_schema()
+
+    # Read config value using existing infrastructure (with error handling)
+    config_value: dict[str, Any] | None = None
+    try:
+        config_value = _get_resource_config_data(
+            resource_config.config_file, resource_config.path_selector
+        )
+    except (OSError, ValueError):
+        pass  # File not readable or path selector invalid
```
AstraBert: Wouldn't we want to throw an error here? it technically means that the config is invalid and the workflow would likely crash

packages/llama-index-workflows/src/workflows/representation/build.py
```diff
@@ -139,21 +185,101 @@ def get_workflow_representation(workflow: Workflow) -> WorkflowGraph:
...
     edges: list[WorkflowGraphEdge] = []
     added_nodes: set[str] = set()  # Track added node IDs to avoid duplicates
-    added_resource_nodes: dict[int, WorkflowResourceNode] = {}  # Track by factory id
-
-    step_config: StepConfig | None = None
+    # Track resource nodes by identity (factory id for _Resource)
+    added_resource_nodes: dict[int, WorkflowResourceNode] = {}
+    # Track resource config nodes by (config_file, path_selector)
+    added_resource_config_nodes: dict[
+        tuple[str, str | None], WorkflowResourceConfigNode
```
AstraBert: I think it would be even easier to have a str as key, by concatenating the config file path and the path selector (you would end up with something like `config_path + ("." + path_selector if path_selector is not None else "")`

AstraBert: Non-blocking tho, just personal preference

packages/llama-index-workflows/tests/test_resources.py
```diff
@@ -506,6 +507,17 @@ def use_repo(
...
+    resource_nodes = [node for node in graph.nodes if node.node_type == "resource"]
+    assert len(resource_nodes) == 2
+    resource_edges = [
+        edge
+        for edge in graph.edges
+        if edge.source.startswith("resource_") and edge.target.startswith("resource_")
+    ]
+    assert len(resource_edges) == 1
+    assert resource_edges[0].label == "db"
+
```
AstraBert: Isn't it technically a duplicate of the testing we do above for the representation module? Just flagging, but I would be fine with having it here

adrianlyjak: ah thanks. I was going back and forth on branches to debug issues with `from __future__ import annotations`. This one looks pretty duplicative, will delete.

packages/llama-index-workflows/src/workflows/representation/build.py
adrianlyjak: fair. I waffled here and let claude do its paranoid thing since there was already a fair amount of paranoid exception handling. Will clean it all out

PR #295, opened 2026-01-22
feat: Make resource dependencies more flexible

adrianlyjak:
- Resources can now be injected into other Resources (recursively, with cycle detection)
- ResourceConfigs can now be injected anywhere a Resource could
- Better handle string / deferred types, usually in association with `from __future__ import annotations`, e.g. `"Annotated[Foo, Resource(get_foo)]"` vs `Annotated[Foo, Resource(get_foo)]`

### DIFF
--- a/packages/llama-index-workflows/src/workflows/representation/build.py
+++ b/packages/llama-index-workflows/src/workflows/representation/build.py
@@ -36,13 +36,14 @@ def _get_resource_identity(resource: ResourceDescriptor) -> int:
     """Get a unique identifier for resource deduplication.
 
     For _Resource, uses the factory function identity.
-    For _ResourceConfig, uses the object identity.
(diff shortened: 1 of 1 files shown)


packages/llama-index-workflows/src/workflows/representation/build.py
```diff
@@ -20,13 +20,32 @@
...
+    """Get a unique identifier for resource deduplication.
+
+    For _Resource, uses the factory function identity.
+    For _ResourceConfig, uses the object identity.
+    """
+    if isinstance(resource, _Resource):
+        return id(resource._factory)
+    if isinstance(resource, _ResourceConfig):
+        # For _ResourceConfig, use the object id or a hash of config file + path_selector
+        return id(resource)
```
AstraBert: Where is the "hash of config file + path selector" option here?

PR #296, opened 2026-01-23
refact: introduce runtime plugin architecture and context separation

adrianlyjak:
- Add Runtime abstract base class with BasicRuntime as default implementation
- Split context into external, internal, and pre-context modules
- Replace workflow_registry with workflow_tracker for workflow lifecycle management
- Remove broker.py in favor of runtime-managed execution
- Add plugins package with get_current_runtime() for context-aware runtime access
- Update handler and workflow to use new runtime interfaces

---

<a href="https://app.devin.ai/review/run-llama/workflows-py/pull/296">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/llama-index-workflows/src/workflows/context/context.py
+++ b/packages/llama-index-workflows/src/workflows/context/context.py
@@ -488,7 +488,10 @@ class Context(Generic[MODEL_T]):
         elif isinstance(self._face, ExternalContext):
             self._face.send_event(message, step)
         else:
-            _warn_send_event_before_start()
+            raise ContextStateError(
+                "send_event() called before workflow started. "
+                "Call workflow.run() first."
+            )
 
     async def wait_for_event(
         self,
@@ -675,14 +678,5 @@ def _warn_cancel_before_start() -> None:
 def _warn_cancel_in_step() -> None:
     warnings.warn(
         "cancel() called from within a step; use send_event() instead.",
-        stacklevel=3,
-    )
-
-
-@functools.lru_cache(maxsize=1)
-def _warn_send_event_before_start() -> None:
-    warnings.warn(
-        "send_event() called before workflow started; event will be dropped. "
-        "Call workflow.run() first.",
         stacklevel=3,
     )
--- a/packages/llama-index-workflows/tests/context/test_context.py
+++ b/packages/llama-index-workflows/tests/context/test_context.py
@@ -15,7 +15,6 @@ from workflows.context.context import (
     _warn_cancel_in_step,
     _warn_get_result,
     _warn_is_running_in_step,
-    _warn_send_event_before_start,
 )
 from workflows.context.external_context import ExternalContext
 from workflows.context.state_store import DictState, InMemoryStateStore
@@ -590,13 +589,12 @@ async def test_cancel_before_start_warns(workflow: Workflow) -> None:
 
 
 @pytest.mark.asyncio
-async def test_send_event_before_start_warns(workflow: Workflow) -> None:
-    """Sending event before run() should emit warning about dropped event."""
-    # Clear the lru_cache to ensure warning fires
-    _warn_send_event_before_start.cache_clear()
-
+async def test_send_event_before_start_raises(workflow: Workflow) -> None:
+    """Sending event before run() should raise ContextStateError."""
     ctx = Context(workflow)
-    with pytest.warns(UserWarning, match="send_event.*called before workflow started"):
+    with pytest.raises(
+        ContextStateError, match="send_event.*called before workflow started"
+    ):
         ctx.send_event(Event())
 
 


packages/llama-index-workflows/src/workflows/resource.py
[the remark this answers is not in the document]
adrianlyjak: Good feedback, but seems kind of unrelated to changes here. Addressed elsewhere

packages/llama-index-workflows/src/workflows/context/external_context.py
```diff
@@ -0,0 +1,171 @@
...
+        self._workflow = workflow
+        self._external_adapter = external_adapter
+        self._serializer = serializer
+        self._workers = []
+
+    @property
+    def is_running(self) -> bool:
+        """Whether the workflow is currently running."""
+        return self._require_v2_runtime_compatibility().is_running
+
```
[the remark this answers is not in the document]
adrianlyjak: Made this a bit more compatible

packages/llama-index-workflows/src/workflows/context/context.py
```diff
@@ -120,163 +125,194 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
 
-    # Backing state store; serialized as `state`
-    _state_store: InMemoryStateStore[MODEL_T]
-    _broker_run: WorkflowBroker[MODEL_T] | None
-    _plugin: Plugin
-    _workflow: Workflow
+    # Current face - context is in exactly one state at a time
+    _face: (
+        PreContext[MODEL_T] | ExternalContext[MODEL_T, Any] | InternalContext[MODEL_T]
+    )
```
logan-markewich, line 131: It feels kind of weird to have a "context" that changes shape depending on the current state of the workflow

Does it make more sense to just introduce 3 attributes/classes with what they need, instead of 1 changing face?

adrianlyjak: Yes, but that would be a more breaking change, so leaving it in this in-between state until v3

packages/llama-index-workflows/src/workflows/runtime/workflow_tracker.py
```diff
@@ -0,0 +1,75 @@
...
+# Copyright (c) 2026 LlamaIndex Inc.
+"""Workflow tracker for runtime instance registration."""
+
+from __future__ import annotations
+
+from workflows.runtime.types.plugin import RegisteredWorkflow
+from workflows.workflow import Workflow
+
+
+class WorkflowTracker:
```
adrianlyjak, line 11: TODO: look into memory leaks in basic runtime

packages/llama-index-workflows/src/workflows/resource.py
```diff
@@ -161,10 +161,19 @@ def get_dependencies(self) -> list[tuple[str, ResourceDescriptor, type | None]]:
...
+) -> dict[str, Any]:
+    # Resolve to absolute path for cache key to handle different working directories
+    abs_path = str(Path(config_file).resolve())
+    return _get_resource_config_data_cached(abs_path, path_selector)
+
+
+@functools.lru_cache(maxsize=128)
+def _get_resource_config_data_cached(
+    config_file: str,
+    path_selector: str | None,
```
adrianlyjak, line 176: unrelated fix for flaky concurrent tests getting in a weird state caching configs from different directories

packages/llama-index-utils-workflow/src/llama_index/utils/workflow/__init__.py
```diff
@@ -463,10 +464,12 @@ def _extract_execution_graph(
...
     """Helper to extract nodes and edges from the workflow handler's tick log."""
-    if handler.ctx is None or handler.ctx._broker_run is None:
-        raise ValueError("No context/run info in this handler. Has it been run yet?")
 
-    ticks: List[WorkflowTick] = handler.ctx._broker_run._tick_log
+    ticks: List[WorkflowTick] = []
+    if handler.ctx is not None:
+        face = handler.ctx._face
+        if isinstance(face, ExternalContext):
+            ticks = face._tick_log
```
logan-markewich, line 471: oof, still not a fan of this face approach (I don't have a better suggestion.... yet)

packages/llama-index-workflows/tests/context/test_context_preservation.py
```diff
@@ -0,0 +1,126 @@
...
+
+See: thoughts/shared/plans/2026-01-23-context-state-preservation-fix.md
+"""
+
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+from workflows.context import Context
```
logan-markewich, line 18: It seems like at least for a user, they are never exposed to the multiple face thing, they still just see a single context?

packages/llama-index-workflows/src/workflows/context/context.py
```diff
@@ -56,6 +60,7 @@ class UnserializableKeyWarning(Warning):
 warnings.simplefilter("once", UnserializableKeyWarning)
 
 
+# TODO(v3) remove this class, and replace with direct references to the pre/internal/external contexts
```
logan-markewich, line 63: I would prefer the context is not deleted? (That'd be pretty disruptive to users). One way to think about it is, to the user, what is the context? Its a gateway to 4 main API's
1. ctx.store
2. ctx.send_event()
3. ctx.cancel_run()
4. ctx.collect_events()

I don't think there's anything else? I wonder if there's a way to maintain an object with these APIs? (Maybe that's the face approach, not sure)

packages/llama-index-workflows/src/workflows/context/context.py
```diff
@@ -120,163 +125,190 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+        if isinstance(self._face, PreContext):
+            return self._face  # type: ignore[invalid-return-type]
+        raise ContextStateError(
+            f"{fn} requires a pre-run context. The workflow has already started."
+        )
+
+    def _require_external(self, fn: str) -> ExternalContext[MODEL_T, Any]:
+        """Require context to be in external state. Raises ContextStateError if not."""
+        if isinstance(self._face, ExternalContext):
+            return self._face
```
logan-markewich, line 180: What is going on here? `fn` is unused?

packages/llama-index-workflows/src/workflows/context/context.py
logan-markewich: 5. ctx.stream_events()

packages/llama-index-workflows/src/workflows/context/context.py
```diff
@@ -120,163 +125,190 @@ async def start(self, ctx: Context, ev: StartEvent) -> StopEvent:
...
+                run_id=run_id,
+                workflow=workflow,
+                init_state=init_state,
+                start_event=start_event,
+                serialized_state=pre.serialized_state,
+                serializer=pre.serializer,
+            )
+
+            # TODO(v3): Remove mutation. Handler will just be the external face.
+            self._face = cast(
```
logan-markewich, line 289: Is there any danger of something going wrong in the lines above before we can change the face? Is that an issue or nah?

packages/llama-index-workflows/src/workflows/context/context.py
```diff
@@ -468,7 +482,13 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> WorkerEvent | GatherEve
...
             ```
         """
-        return self._running_broker.send_event(message, step)
+        # send_event can be called from internal (steps) or external (handler) contexts
+        if isinstance(self._face, InternalContext):
+            self._face.send_event(message, step)
+        elif isinstance(self._face, ExternalContext):
+            self._face.send_event(message, step)
+        else:
+            _warn_send_event_before_start()
```
logan-markewich: Does this remove the ability to call `send_event()` before calling workflow.run()? Or that was already the case before? (Vaguely remembering something about this)

packages/llama-index-workflows/src/workflows/context/context.py
adrianlyjak: I think it would end up more like
- WorkflowHandler + ExternalContext would merge, remove some methods
- Context would remain the internal face, with the same name. This is the one with the type everywhere, so I'd like to keep that easy to migrate, but many methods would go away
- Some other context builder / context artifact would be passed, like `workflow.run(ctx=previous_ctx.continuation)` and/or `workflow.run(ctx=ContextSnapshot.with_state({"foo": "bar"}))`

packages/llama-index-workflows/src/workflows/context/context.py
adrianlyjak: `fn` is just the name of a thing to put in the error message, to give a little bit of a better error message

packages/llama-index-workflows/src/workflows/context/context.py
adrianlyjak: I think you'd need to be passing the same context to multiple runs concurrently which is a weird thing to do

packages/llama-index-workflows/src/workflows/context/context.py
adrianlyjak: I’ll look into it. Definitely removed that functionality if it was present

packages/llama-index-workflows/tests/context/test_context_preservation.py
adrianlyjak: yeah, that's the idea. Just helps model it a bit better

packages/llama-index-workflows/src/workflows/context/context.py
logan-markewich: oh lol

packages/llama-index-workflows/src/workflows/context/context.py
adrianlyjak: Ok, in the last iteration it errored out rather than dropping the message. I'll update to keep this behavior

https://github.com/run-llama/workflows-py/blob/33bbd23/packages/llama-index-workflows/src/workflows/context/context.py#L273-L280

PR #303, opened 2026-01-23
Add resource validation to workflow.validate()

adrianlyjak:
Workflow validate can pass on something that will then crash later on from failing to load e.g. Resources/ResourceConfigs. By default, validation now loads the configs and validates them as well, and you can opt in to fully resolving Resources (e.g. clients). That one is disabled by default, as environment variables may not be set up.

### DIFF
--- a/packages/llama-index-workflows/tests/test_resources.py
+++ b/packages/llama-index-workflows/tests/test_resources.py
@@ -947,19 +947,6 @@ def test_validate_resource_factory_failure() -> None:
         wf.validate(validate_resources=True)
 
 
-def test_validate_without_resources() -> None:
-    """Test that validation works fine for workflows without resources."""
-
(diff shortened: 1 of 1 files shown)


packages/llama-index-workflows/tests/test_resources.py
```diff
@@ -701,3 +701,288 @@ def use_db_config(
...
+    """Test that validation works fine for workflows without resources."""
+
+    class TestWorkflow(Workflow):
+        @step
+        def start_step(self, ev: StartEvent) -> StopEvent:
+            return StopEvent()
+
+    wf = TestWorkflow(disable_validation=True)
+    # Should not raise
+    wf.validate()
```
AstraBert: I'm sure this might be tested elsewhere so this is potentially a duplicate

PR #315, opened 2026-01-30
Remove asyncio queue from control_loop

adrianlyjak:
Splitting off some changes from DBOS work

### Control Loop Changes
- Switches out the asyncio delay mechanism for a pull-with-timeout that is more deterministic friendly
- Adds a priority queue of delayed tasks
- Switches out the misc firing /spawning of async tasks to a more rigorous pattern where tasks are only created in the main loop, and gathered in one location. This makes the concurrency more straightforward to reason about
- Moves the "selecting" of which tasks is completed first into an adapter responsibility, this allows more granular journaling decisions (currently needed to make DBOS determinism work. We end up keeping our own lightweight journal as a workaround until we can figure out a better way to integration)
- adds a named task around the task to make journaling easier

### Misc Changs
- Removes mypy. 3 type checkers is too many
- Starts a place for runtime matrix tests. I put these in the integration tests as I expect they will frequently be slow
- Reduces some test warning changes
- Fix bug where run_id is not actually passed through from workflow.run

---

<a href="https://app.devin.ai/review/run-llama/workflows-py/pull/315">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/llama-index-workflows/src/workflows/runtime/control_loop.py
+++ b/packages/llama-index-workflows/src/workflows/runtime/control_loop.py
@@ -74,14 +74,6 @@ from workflows.runtime.types.ticks import (
 from workflows.workflow import Workflow
 
 
-class _DelayedTick:
-    """Sentinel class representing a tick that should be scheduled after a delay."""
-
-    def __init__(self, tick: WorkflowTick, delay: float) -> None:
-        self.tick = tick
-        self.delay = delay
-
-
 async def _single_pull(adapter: InternalRunAdapter) -> WorkflowTick | None:
     """Single-iteration pull: calls wait_receive once and returns the tick.
 
@@ -131,8 +123,11 @@ class _ControlLoopRunner:
         # Pending items to be processed (from rehydration or delayed ticks)
         for tick in self.state.rehydrate_with_ticks():
             self.tick_buffer.append(tick)
-        # Scheduled wakeups: heap of (wakeup_time, tick) tuples
-        self.scheduled_wakeups: list[tuple[float, WorkflowTick]] = []
+        # Scheduled wakeups: heap of (wakeup_time, sequence, tick) tuples
+        # The sequence counter ensures deterministic ordering when timestamps are equal,
+        # avoiding TypeError from comparing WorkflowTick objects that don't implement __lt__
+        self.scheduled_wakeups: list[tuple[float, int, WorkflowTick]] = []
+        self._wakeup_sequence = 0
         self.snapshot_adapter = as_snapshottable_adapter(adapter)
         # Pull task sequence counter for deterministic journaling
         self._pull_sequence = 0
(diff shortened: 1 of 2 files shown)


packages/llama-agents-server/src/llama_agents/server/server.py
```diff
@@ -1762,33 +1763,41 @@ async def _stream_events(self, on_finish: Callable[[], Awaitable[None]]) -> None
...
-                self.queue.put_nowait(event)
+                        if state is None:
+                            logger.warning(
+                                f"Context state is None for handler {self.handler_id}. This is not expected."
+                            )
+                            continue
+                        await self.checkpoint()
+
+                    self.queue.put_nowait(event)
+            except WorkflowRuntimeError:
```
logan-markewich, line 1796: Is "stream already consumed" the only error that raises WorkflowRuntimeError?

packages/llama-index-workflows/src/workflows/context/state_store.py
```diff
@@ -23,6 +33,165 @@
...
+        state_data = serializer.serialize(state)
+
+    return InMemorySerializedState(
+        state_type=type(state).__name__,
+        state_module=type(state).__module__,
+        state_data=state_data,
+    )
+
+
+def traverse_path_step(obj: Any, segment: str) -> Any:
```
logan-markewich, line 144: I honestly forgot this existed -- is it still used?

logan-markewich: it is! neat

packages/llama-index-workflows/src/workflows/plugins/basic.py
```diff
@@ -91,28 +101,39 @@ def run_id(self) -> str:
...
+            if timeout_seconds is None:
+                tick = await self._queues.receive_queue.get()
+            else:
+                tick = await asyncio.wait_for(
+                    self._queues.receive_queue.get(),
+                    timeout=timeout_seconds,
+                )
+            return WaitResultTick(tick=tick)
+        except asyncio.TimeoutError:
+            return WaitResultTimeout()
```
logan-markewich, line 128: whats the intended behaviour when a tick times out?

logan-markewich (approved):
I think this is ok... slightly worried that the code is getting extremely complex, so I hope its still debuggable (as compared to v1 which was just asyncio tasks in a single gather). I'm slowly losing grip on my understanding of the codebase lol I may need to spend more time with it.

Devin raised some valid remaining points as well that should be addressed

adrianlyjak:
> I think this is ok... slightly worried that the code is getting extremely complex, so I hope its still debuggable (as compared to v1 which was just asyncio tasks in a single gather). I'm slowly losing grip on my understanding of the codebase lol I may need to spend more time with it.

I think we're doing pretty good. control_loop, which is the core of the logic is ~1000k lines, basic adapter is ~300. Other stuff is types and a bunch of API stability. In general this should help a lot. There's a lot of race conditions removed, and we could fairly easily consider alternate storage mechanisms for the basic runtime, or sort of timeline debug and time-travel features built on top.

> Devin raised some valid remaining points as well that should be addressed

Already fixed!

PR #316, opened 2026-01-31
Type with StateStore protocol rather than concrete InMemoryStateStore

adrianlyjak:
To facilitate later state stores, switch out InMemoryStateStore for StateStore protocol with same method interface. 

This might cause some typing churn in clients, but is otherwise identical in behavior


logan-markewich (approved):
Seems safe to me!

PR #340, opened 2026-02-06
Fix concurrent step cancellation regression in 2.14.0

### LINKED ISSUE #339: Concurrent step not cancelled after StopEvent in 2.14.0
After upgrading from **llama-index-workflows 2.13.1** to **2.14.0**, one of my workflow tests started failing.

The workflow has two concurrent steps. The test covers the case where one step finishes early and returns a StopEvent, which should cancel the other concurrent step.

#### Expected behavior (2.13.1)

```
StartEvent -> some_step       -> StopEvent
           -> some_other_step -X (cancelled)
```

Event stream:

```
[StopEvent]
```

#### Actual behavior (2.14.0)

```
StartEvent -> some_step       -> StopEvent
           -> some_other_step -------------> OtherEvent
```

Event stream:

```
[OtherEvent, StopEvent]
```

It looks like `StopEvent` no longer cancels the running concurrent steps in 2.14.0.

adrianlyjak:
## Summary

Fixes #339

The 2.14.0 refactor from `asyncio.Queue` to `tick_buffer` + `wait_for_next_task` changed cancellation timing so that a `StopEvent` returned by one step no longer cancelled sibling concurrent steps before they could write to the event stream or have their results processed. This widened the window where a race between a `StopEvent` and a concurrent step's stream write could occur — races are still theoretically possible due to concurrency, but in 2.14.0 even a single `asyncio.sleep(0)` yield was enough for the write to slip through, whereas previously it was not.

This PR fixes the regression by eagerly cancelling worker tasks when a `StopEvent` is detected:
- When a worker returns a `StopEvent`, immediately call `cleanup_tasks()` to cancel siblings before buffering the tick result
- When `CommandCompleteRun` is processed, call `cleanup_tasks()` to ensure all workers are stopped before returning

### DIFF
--- a/.changeset/fix-concurrent-step-cancellation.md
+++ b/.changeset/fix-concurrent-step-cancellation.md
@@ -2,4 +2,4 @@
 "llama-index-workflows": patch
 ---
 
-Fix concurrent step cancellation regression where StopEvent no longer cancelled sibling steps
+Fix concurrent step cancellation regression where StopEvent no longer cancelled as quickly as previously


logan-markewich (approved):
Easy fix!

PR #352, opened 2026-02-09
Add AgentDataStore backed by LlamaCloud Agent Data API

adrianlyjak:
Add a simple workflow store backed by llama cloud agent data. This moves it out the of the core of the llama-deploy-appserver to be more readily testable, and integrates it with the new runtime protocol

### DIFF
--- a/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_state_store.py
+++ b/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_state_store.py
@@ -129,10 +129,10 @@ class AgentDataStateStore(Generic[MODEL_T]):
             return self._deserialize_state(record.data)
         return None
 
-    async def _save_state(self, state: MODEL_T) -> None:
+    async def _save_state(self, state: BaseModel) -> None:
         record = _StoredStateRecord(
             run_id=self._run_id,
-            data=self._serialize_state(state),
+            data=self._serialize_state(state),  # type: ignore[arg-type]
         )
         payload = record.model_dump()
         if self._item_id is not None:
@@ -165,7 +165,7 @@ class AgentDataStateStore(Generic[MODEL_T]):
             await self._save_state(state)
             return
         merged = merge_state(current, state)
-        await self._save_state(merged)  # type: ignore[arg-type]
+        await self._save_state(merged)
 
     async def get(self, path: str, default: Any = ...) -> Any:
         state = await self._load_state()
--- a/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
+++ b/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
@@ -196,13 +196,17 @@ class AgentDataStore(AbstractWorkflowStore):
         if filters is None:
             return 0
 
+        # Invalidate cached IDs for matching handlers before bulk delete
         items = await self._client.search(self._collection, filters or None)
         for item in items:
             handler_id = item["data"].get("handler_id")
             if handler_id:
                 self._id_cache.delete(handler_id)
-            await self._client.delete_item(item["id"])
-        return len(items)
+
+        if not items:
+            return 0
+
+        return await self._client.delete_many(self._collection, filters or {})
 
     # ------------------------------------------------------------------
     # Event journal
--- a/packages/llama-agents-integration-tests/src/llama_agents_integration_tests/fake_agent_data.py
+++ b/packages/llama-agents-integration-tests/src/llama_agents_integration_tests/fake_agent_data.py
@@ -101,6 +101,18 @@ class FakeAgentDataBackend:
                     return
         raise ValueError(f"Item {item_id} not found")
 
+    def delete_many(
+        self,
+        deployment_name: str,
+        collection: str,
+        filters: dict[str, Any],
+    ) -> int:
+        items = self._get_items(deployment_name, collection)
+        matched = [item for item in items if self._matches(item["data"], filters)]
+        for item in matched:
+            items.remove(item)
+        return len(matched)
+
     def handle_request(self, request: httpx.Request) -> httpx.Response:
         """Route an httpx.Request to the appropriate handler."""
         path = request.url.path
@@ -116,6 +128,15 @@ class FakeAgentDataBackend:
                 body.get("order_by"),
             )
             return httpx.Response(200, json={"items": items})
+
+        if method == "POST" and path == "/api/v1/beta/agent-data/:delete":
+            body = json.loads(request.content)
+            count = self.delete_many(
+                body["deployment_name"],
+                body["collection"],
+                body.get("filter", {}),
+            )
+            return httpx.Response(200, json={"deleted_count": count})
 
         if method == "POST" and path == "/api/v1/beta/agent-data":
             body = json.loads(request.content)
--- a/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_client.py
+++ b/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_client.py
@@ -101,3 +101,19 @@ class AgentDataClient:
         async with self.http_client() as client:
             resp = await client.delete(f"/api/v1/beta/agent-data/{item_id}")
             resp.raise_for_status()
+
+    async def delete_many(
+        self,
+        collection: str,
+        filters: dict[str, Any],
+    ) -> int:
+        """Delete items matching the given filters. Returns the number deleted."""
+        body: dict[str, Any] = {
+            "deployment_name": self._deployment_name,
+            "collection": collection,
+            "filter": filters,
+        }
+        async with self.http_client() as client:
+            resp = await client.post("/api/v1/beta/agent-data/:delete", json=body)
+            resp.raise_for_status()
+            return resp.json().get("deleted_count", 0)


packages/llama-agents-integration-tests/tests/test_server_store_matrix.py
```diff
@@ -0,0 +1,529 @@
...
+
+
+async def _get_handler_raw(base_url: str, handler_id: str) -> dict[str, Any]:
+    """Get handler data without raising on non-200 status (server returns 500 for failed/cancelled)."""
+    async with httpx.AsyncClient(base_url=base_url, timeout=5.0) as client:
+        resp = await client.get(f"/handlers/{handler_id}")
+        return resp.json()  # type: ignore[no-any-return]
+
+
+async def wait_for_passing(
```
adrianlyjak: if there's an existing utility like this already, this should be replaced, otherwise this should be moved to somewhere shared

packages/llama-agents-integration-tests/tests/test_server_store_matrix.py
```diff
@@ -0,0 +1,529 @@
...
+
+@pytest.mark.asyncio
+async def test_list_handlers(
+    server_with_store: tuple[str, WorkflowServer, str],
+) -> None:
+    base_url, _server, _store_type = server_with_store
+    client = WorkflowClient(base_url=base_url)
+
+    # Run two workflows
+    await client.run_workflow("test", start_event={"__is_pydantic_event": True})
```
adrianlyjak: I think we can just use `StartEvent()`

packages/llama-agents-integration-tests/tests/test_server_store_matrix.py
```diff
@@ -0,0 +1,529 @@
...
+@pytest.mark.asyncio
+async def test_cursor_resume(
+    server_with_store: tuple[str, WorkflowServer, str],
+) -> None:
+    base_url, _server, _store_type = server_with_store
+    client = WorkflowClient(base_url=base_url)
+
+    # Run a streaming workflow synchronously first so all events are stored
+    result = await client.run_workflow(
+        "streaming", start_event={"__is_pydantic_event": True}
```
adrianlyjak: ditto StartEvent

packages/llama-agents-integration-tests/tests/test_server_store_matrix.py
```diff
@@ -0,0 +1,529 @@
...
+    )
+    assert result.status == "completed"
+    handler_id = result.handler_id
+
+    # Now fetch events with after_sequence to get only later events
+    async with httpx.AsyncClient(base_url=base_url, timeout=5.0) as http_client:
```
adrianlyjak: doesn't the client support specifying an after sequence? Odd that we're excercising the rarer sse=false path

packages/llama-agents-integration-tests/tests/test_server_store_matrix.py
```diff
@@ -0,0 +1,529 @@
...
+
+    async def all_completed() -> None:
+        for hid in handler_ids:
+            data = await client.get_handler(hid)
+            assert data.status == "completed"
+
+    await wait_for_passing(all_completed)
+
+
+# -- Durability tests (sqlite only) --
```
adrianlyjak: wait, lots of these are durable. This should just not excercise the in memory path. The others should parameterize for durable tests

packages/llama-agents-server/src/llama_agents/server/_store/sqlite/sqlite_workflow_store.py
```diff
@@ -212,22 +208,22 @@ async def subscribe_events(
...
-            events = await self.query_events(run_id, after_sequence=cursor)
-            for event in events:
+            async with condition:
+                batch = await self.query_events(run_id, after_sequence=cursor)
+                if not batch:
+                    with contextlib.suppress(TimeoutError):
+                        await asyncio.wait_for(
+                            condition.wait(), timeout=self.poll_interval
+                        )
+                    batch = await self.query_events(run_id, after_sequence=cursor)
```
adrianlyjak: wait, this looks like a double query. Shouldn't this not be here, and we just query first, and if none, wait?

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_state_store.py
```diff
@@ -0,0 +1,200 @@
...
+            return self._deserialize_state(items[0]["data"]["state_json"])
+        return None
+
+    async def _save_state(self, state: MODEL_T) -> None:
+        state_json = self._serialize_state(state)
+        data = {
+            "run_id": self._run_id,
+            "state_json": state_json,
+            "state_type": type(state).__name__,
+            "state_module": type(state).__module__,
```
adrianlyjak: the type / module seem unnecessary and dangerous. We should be interpreting that from the code context, not from the serialized data. A lot of this is likely duplicated in _serialize_state. This is WAY too nested. We just want run_id: xyz, data: nice_simple_json

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_state_store.py
```diff
@@ -0,0 +1,200 @@
...
+        await self._save_state(state)
+        return state
+
+    async def _load_state_or_none(self) -> MODEL_T | None:
+        items = await self._client.search(
+            self._collection, {"run_id": {"eq": self._run_id}}, page_size=1
+        )
+        if items:
+            self._item_id = items[0]["id"]
+            return self._deserialize_state(items[0]["data"]["state_json"])
```
adrianlyjak: see later comment. This should just be items[0]["data"]. Better yet, this should have a private intermediary model to describe / validate this structure

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_state_store.py
```diff
@@ -0,0 +1,200 @@
...
+        state = self._create_default_state()
+        await self._save_state(state)
+        return state
+
+    async def _load_state_or_none(self) -> MODEL_T | None:
+        items = await self._client.search(
+            self._collection, {"run_id": {"eq": self._run_id}}, page_size=1
```
adrianlyjak: we should have constants for run_id/data

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_state_store.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        if items:
+            self._item_id = items[0]["id"]
+            return self._deserialize_state(items[0]["data"]["state_json"])
+        return None
+
+    async def _save_state(self, state: MODEL_T) -> None:
```
adrianlyjak: this should likely not be MODEL_T, considering we send it the merged model? At this point, a dict is probably appropriate in this private method

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+    # ------------------------------------------------------------------
+    # Handler CRUD
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def _build_handler_filters(query: HandlerQuery) -> dict[str, Any] | None:
+        """Convert a HandlerQuery to Agent Data API filter format."""
+        filters: dict[str, Any] = {}
+
+        if query.handler_id_in is not None:
```
adrianlyjak: this is highly repetitive. We should make a helper to facilitate

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+            if len(query.handler_id_in) == 0:
+                return None  # empty list → match nothing
+            if len(query.handler_id_in) == 1:
+                filters["handler_id"] = {"eq": query.handler_id_in[0]}
```
adrianlyjak: this doesn't look useful. Just always use an includes query

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+
+    @staticmethod
+    def _item_to_handler(item: dict[str, Any]) -> PersistentHandler:
+        """Convert an Agent Data API item to a PersistentHandler."""
+        data = item["data"]
+        return PersistentHandler.model_validate(data)
+
+    async def query(self, query: HandlerQuery) -> List[PersistentHandler]:
+        filters = self._build_handler_filters(query)
+        # None means an empty filter list was given → match nothing
```
adrianlyjak: this feels like we should have different return types to indicate "really query for everything" vs "no, don't query anything". This seems like a potential large footgun to have this after check. Also, why don't we just always return filters, even if empty? That seems the natural solution. None means gtfo, {} means "I mean everything, really"

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+            )
+            if items:
+                item_id = items[0]["id"]
+                self._id_cache.put(handler_id, item_id)
+                await self._client.update_item(item_id, data)
+            else:
+                result = await self._client.create(self._collection, data)
+                self._id_cache.put(handler_id, result["id"])
+
+    async def delete(self, query: HandlerQuery) -> int:
```
adrianlyjak, line 194: wait, this is weird. There's a delete many api. Seems like we should manage this more like 
query->invalidate cache->delete batch. Right now this seems pretty inefficient

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+            timestamp=now,
+            event=event,
+        )
+        await self._client.create(
+            self._events_collection,
+            stored.model_dump(mode="json"),
+        )
+        # Notify subscribers
+        condition = self._conditions.get(run_id)
+        if condition is not None:
```
adrianlyjak: nit: `if condition := self._conditions.get(run_id)`

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+        self,
+        run_id: str,
+        after_sequence: int | None = None,
+        limit: int | None = None,
+    ) -> list[StoredEvent]:
+        filters: dict[str, Any] = {"run_id": {"eq": run_id}}
+        if after_sequence is not None:
+            filters["sequence"] = {"gte": after_sequence + 1}
+
+        items = await self._client.search(
```
adrianlyjak, line 252: erm, client side sort with a max? There's a `order_by` param in the API. 

```
Filtering/Sorting: You can filter on any data fields and on the top‑level created_at and updated_at. Sorting accepts a comma‑separated list; prefix fields inside data with data. (for example, "data.name desc, created_at").
```

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_store.py
```diff
@@ -0,0 +1,363 @@
...
+            stored.model_dump(mode="json"),
+        )
+
+    async def get_ticks(self, run_id: str) -> list[StoredTick]:
+        items = await self._client.search(
+            self._ticks_collection,
+            {"run_id": {"eq": run_id}},
+        )
+        ticks = [StoredTick.model_validate(item["data"]) for item in items]
+        ticks.sort(key=lambda t: t.sequence)
```
adrianlyjak: this is terrible. Everything will start breaking down after the page limit is hit. Also, this needs to paginate all items, and ought to use the api sort

packages/llama-agents-server/src/llama_agents/server/_store/memory_workflow_store.py
```diff
@@ -177,34 +172,29 @@ async def subscribe_events(
...
+
         while True:
-            all_events = self.events.get(run_id, [])
-            batch = all_events[cursor:]
+            async with condition:
+                all_events = self.events.get(run_id, [])
+                batch = all_events[cursor:]
+                if not batch:
+                    await condition.wait()
+                    all_events = self.events.get(run_id, [])
```
adrianlyjak: unnecessary double query

packages/llama-agents-server/src/llama_agents/server/_store/postgres_workflow_store.py
```diff
@@ -359,22 +357,28 @@ async def subscribe_events(
...
+                )
+                batch = [e for e in events if e.sequence > cursor]
+                if not batch:
+                    with contextlib.suppress(TimeoutError):
                         await asyncio.wait_for(
                             condition.wait(), timeout=self.poll_interval
                         )
-                    except TimeoutError:
-                        pass
+                    events = await self.query_events(
```
adrianlyjak: unnecessary double query

Issue #380, opened 2026-02-13
Workflow step spans are no longer parented under workflow run span since v2.14.0


schelv:
Starting from `llama-index-workflows==2.14.0`, workflow step spans are no longer grouped under their corresponding workflow run span in Arize Phoenix.

In `v2.13.1`, all step spans are correctly nested under their corresponding workflow run span (see screenshot, bottom section).
In `v2.14.0`, the step spans appear as separate top-level spans instead of being parented to the workflow run (see screenshot, top section).

<img width="1072" height="565" alt="Image" src="https://github.com/user-attachments/assets/bf7b7e2e-1fab-46d8-a530-e84503abe251" />

Tracing is still emitted, but span parent relationships appear to be broken.

In `2.13.1`, the `Context` object memory address (`0x...`) remains constant across workflow steps. In `2.14.0`, a new `Context` object appears per step (different `0x...`), which may indicate a change in context propagation.

**Reproduction**

* Works as expected in: `2.13.1`
* Broken in: `2.14.0`
* Observed in Arize Phoenix

PR #381, opened 2026-02-13
fix span handling and span hierarchy propagation

logan-markewich:
The switch from workflow handler being an asyncio future to an awaitable caused a few bugs
- async context being lost
- the llama-index-instrumentation code not being setup to handle `Awaitable` types

Fixes: https://github.com/run-llama/workflows-py/issues/380

We could move awaitable handling into the instrumentation package, but I'd rather ensure its fixed here first

### DIFF
--- a/.changeset/afraid-terms-kick.md
+++ b/.changeset/afraid-terms-kick.md
@@ -1,0 +1,5 @@
+---
+"llama-index-workflows": patch
+---
+
+Fix span tracking in observability tooling


packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -410,25 +414,92 @@ def run(
...
+            start_event_instance: StartEvent | None = (
+                None
+                if ctx.is_running
+                else self._get_start_event_instance(start_event, **kwargs)
+            )
+            handler = ctx._workflow_run(
+                workflow=self, start_event=start_event_instance, run_id=run_id
+            )
+
+            # Add callback to close span when workflow completes
```
adrianlyjak, line 463: This works as a bandaid, but this may be a little brittle at this outer layer. There are starting to be patterns for fetching a workflow handler on demand, given a run id, to interface with it as a client. It also makes sense to pass this through to something a little more in the core--Not sure if this is going to work on dbos steps. 

- The handoff of tags and span ids kind of needs to be serialized into some context that the step starts from to properly attribute it (then you wouldn't need the code direct in the basic plugin)
- starting / completing spans makes sense to keep close to the core in the control loop, as that's what really knows about started/completed

logan-markewich: As soon as steps are distributed/disconnected from the initial runtime (like dbos), I fully expect all tracing to break

logan-markewich: Distributed tracing is not on my bingo card for now :p

PR #386, opened 2026-02-19
add dbos idle

adrianlyjak:


AGENTS.md
```diff
@@ -30,6 +30,12 @@ uv run dev -- -k test_name
 
 For more advanced scenarios, you can always `cd packages/some-package` and use pytest directly. The dev tool just provides additional package level test parallelism, and more curated cross package test output to avoid context bloat.
 
+The `llama-agents-dbos` package has Docker integration tests (requires Docker running) marked with `@pytest.mark.docker`. Run them with:
```
adrianlyjak: No, this is other packages too

packages/llama-agents-dbos/tests/test_dbos_server_postgres.py
```diff
@@ -182,3 +191,243 @@ def test_no_duplicate_events_after_replay(postgres_dsn: str) -> None:
         assert int(streams_count) == 0, (
             f"Expected 0 events in dbos.streams after replay, got {streams_count}"
         )
+
+
+def run_cancel_resume_scenario(
```
adrianlyjak: There's way too many of these e2e tests. Tone it down. Which tests actually test the full end to end scenario, while also validating that underlying threads are released. Probably just needs one good test

packages/llama-agents-dbos/tests/test_dbos_cross_process.py
```diff
@@ -0,0 +1,120 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+"""Cross-process integration tests for DBOS distributed model.
```
adrianlyjak, line 3: this is good, I wasn't referring to this test with the e2e test comment. Needed coverage here

packages/llama-agents-dbos/tests/fixtures/cancel_hangs_runner.py
```diff
@@ -0,0 +1,157 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+"""Test runner: does idle-release cancel actually unblock the control loop?
+
+Verifies that after the idle release decorator cancels a workflow, the control
+loop task (blocked in recv_async) actually exits — not just that the DBOS status
```
adrianlyjak: this should just be validated integrated into other tests

packages/llama-agents-dbos/src/llama_agents/dbos/idle_release.py
```diff
@@ -0,0 +1,273 @@
...
+        # The original workflow's recv consumed its notification live (DELETE),
+        # but DBOS recv replay returns cached results WITHOUT deleting from DB.
+        # Any notifications inserted after the original recv (e.g. the wake-up
+        # None from _release_idle_handler) persist and would be consumed by the
+        # resumed workflow's fresh recv instead of the real tick.
+        dbos_inst = _get_dbos_instance()
+        with dbos_inst._sys_db.engine.begin() as conn:
+            conn.execute(
+                sa.text(
+                    "DELETE FROM dbos.notifications"
```
adrianlyjak: this is certainly a hack. Need a better understanding of how notifications work and make this comment clearer if this in any way acceptable, or what forward compatibility would look like for an upstream change around this scenario

packages/llama-agents-dbos/src/llama_agents/dbos/idle_release.py
```diff
@@ -0,0 +1,359 @@
...
+        # Rebuild BrokerState from persisted ticks
+        init_state = await self._broker_state_from_ticks(workflow, run_id)
+
+        # Include the pending tick in the rebuilt state so the control loop
+        # has it queued before it starts processing.
+        if pending_tick is not None:
+            init_state = rebuild_state_from_ticks(init_state, [pending_tick])
+
+        # Carry over state from old run's state store
+        serializer = JsonSerializer()
```
adrianlyjak, line 321: This looks fishy. Should be able to just pass the old to dict

packages/llama-agents-dbos/src/llama_agents/dbos/idle_release.py
```diff
@@ -0,0 +1,359 @@
...
+                if state is not None:
+                    payload = create_in_memory_payload(state, serializer)
+                    serialized_state = payload.model_dump()
+            except Exception:
+                logger.warning(
+                    f"Failed to carry over state from run {run_id}", exc_info=True
+                )
+
+        # Purge DBOS state and journal (if not already purged) so the same run_id can be reused
+        if self._journal_crud is not None:
```
adrianlyjak, line 336: Fishy double delete

packages/llama-agents-dbos/tests/fixtures/sample_workflows/idle_cancel_resume.py
```diff
@@ -0,0 +1,43 @@
...
+
+class AskForDataEvent(InputRequiredEvent):
+    prompt: str = Field(default="Provide data")
+
+
+class ExternalDataEvent(HumanResponseEvent):
+    response: str = Field(default="")
+
+
+class IdleCancelResumeWorkflow(Workflow):
```
adrianlyjak: This should use human in the loop events rather, might be duplicative and should just delete

packages/llama-agents-dbos/tests/fixtures/cancel_hangs_runner.py
adrianlyjak: Additionally, the premise is wrong. This was never testing hanging. Rather it’s validating that cleaning up the workflow actually releases resources. This is now inherent, so it should just be deleted

packages/llama-agents-dbos/tests/fixtures/idle_release_runner.py
```diff
@@ -0,0 +1,178 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+"""Subprocess runner for DBOS idle release end-to-end integration test.
+
```
adrianlyjak: Review differences between this and the idle release http runner. One has got to go, probably this one

packages/llama-agents-dbos/tests/test_dbos_idle_release.py
```diff
@@ -0,0 +1,348 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+"""Unit tests for DBOSIdleReleaseDecorator."""
```
adrianlyjak, line 3: The mocking is too heavy here. There should be light weight real components that can be used instead for some components

packages/llama-agents-dbos/src/llama_agents/dbos/idle_release.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @override
+    def get_external_adapter(self, run_id: str) -> ExternalRunAdapter:
+        return DBOSIdleReleaseExternalRunAdapter(self, run_id)
+
+    async def _deferred_release(self, run_id: str) -> None:
+        """Wait for idle_timeout then release the handler if still idle."""
+        await asyncio.sleep(self._idle_timeout)
+        self._deferred_release_tasks.pop(run_id, None)
+        await self._release_idle_handler(run_id)
```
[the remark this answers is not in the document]
adrianlyjak: This would be actively harmful. If another event arrives simultaneous, it will be journaled before the idle release tick, and replayed during resumption (there's probably a bug in there where it needs another event to wake back up). However if we do cancel this while releasing it would be very harmful, and leave the workflow in an inconsisting (stuck releasing state). 

Perhaps the release event can return data about whether it was actually idle? That could be used to immediately resume as a safeguard

PR #408, opened 2026-03-06
wip: start creating an agentcore memory storage backend

AstraBert:
Create an AgentCore-based storage backend for workflow servers

### DIFF
--- a/packages/llama-agents-server/pyproject.toml
+++ b/packages/llama-agents-server/pyproject.toml
@@ -34,8 +34,7 @@ dependencies = [
 [project.optional-dependencies]
 asyncpg = ["asyncpg>=0.29.0"]
 agentcore = [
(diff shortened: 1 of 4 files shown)


packages/llama-agents-server/pyproject.toml
```diff
@@ -33,6 +33,9 @@ dependencies = [
 
 [project.optional-dependencies]
 asyncpg = ["asyncpg>=0.29.0"]
+agentcore = [
+  "bedrock-agentcore>=0.0.1"
```
adrianlyjak, line 37: let's keep this entirely in the llama-agents-bedrock package, rather than an optional part of the core, similar to DBOS

AstraBert: Closing it here then, will move to `llama-agents-bedrock`

PR #422, opened 2026-03-12
Fix wait_for_event timeout not being enforced

enrico-stauss:
The timeout parameter passed to ctx.wait_for_event() was stored in AddWaiter but never used to schedule a timeout tick, so waiters could hang indefinitely regardless of the configured timeout.

Fix by introducing TickWaiterTimeout and CommandScheduleWaiterTimeout. When AddWaiter is processed, a TickWaiterTimeout is scheduled at now+timeout. When it fires, the waiter is marked timed_out=True and the step is replayed. On replay, wait_for_event detects timed_out and raises asyncio.TimeoutError as documented. A test verifies the behaviour.

### DIFF
--- a/.changeset/neat-months-play.md
+++ b/.changeset/neat-months-play.md
@@ -1,0 +1,5 @@
+---
+"llama-index-workflows": patch
+---
+
+Fix wait_for_event timeout not being enforced


adrianlyjak:
❤️

PR #426, opened 2026-03-14
feat: add graph structure validation

### LINKED ISSUE #415: Enhance validate() with graph-based structural analysis
The current `_validate()` does a solid job checking event production/consumption as flat sets — it catches missing producers, orphaned events, and invalid `StopEvent` usage. This proposes extending it with graph-based structural analysis to catch a class of misconfigurations that currently only surface at runtime.

### Motivation

For complex workflows with many steps and branching event paths, it's possible to have all events properly paired (every produced event is consumed, every consumed event is produced) while the graph itself is structurally broken. These cases pass `validate()` today and only manifest as runtime hangs or unexpected behavior:

1. **Unreachable steps** — a step accepts an event that no transitive path from `StartEvent` ever produces
2. **Dead-end steps** — a step produces events, but no transitive path from it reaches `StopEvent`
3. **No complete path** — events balance in sets, but the directed graph from `StartEvent` to `StopEvent` is disconnected
4. **Unbreakable cycles** — A→B→A with no branch leading to `StopEvent` (workflow hangs until timeout)

### Proposed Approach

Leverage the existing `get_workflow_representation()` infrastructure in `representation/build.py`, which already builds a full `WorkflowGraph` with step nodes, event nodes, and directed edges. Layer graph traversal checks (BFS/DFS reachability, cycle-exit analysis) on top of the graph it already constructs.

### Usage

```python
w = MyWorkflow()
w.validate()  # API unchanged; raises WorkflowValidationError with actionable messages
```

Example error messages:

```
WorkflowValidationError: Step 'process_data' is unreachable from StartEvent.
WorkflowValidationError: Step 'transform' leads to a dead end — no path to StopEvent.
WorkflowValidationError: Cycle detected [step_a → EventX → step_b → EventY → step_a] with no exit path to StopEvent.
```

### Scope

- Extend `_validate()` in `workflow.py` to build the step-event graph via existing `get_workflow_representation()`
- Add reachability, dead-end, path-existence, and cycle-exit checks
- Unit tests covering each misconfiguration type
- No new public API — same `validate()` / `WorkflowValidationError` contract
- No new dependencies — pure graph algorithms on existing `representation/` infrastructure

### Open Questions

Happy to discuss the approach before starting the PR — especially:

- Whether these checks should live inside `_validate()` or as a separate `validate_graph()` method
- Whether any of the four checks should be warnings vs. hard errors
- Whether the graph checks should run on every `run()` call (like current validation) or only on explicit `validate()` calls, given the added traversal cost

vasudev13:
Closes #415. Adds graph-structure validation to Workflow.validate() so misconfigurations are caught before runtime.

Rule 1 — Reachability: Every step must be reachable from an input (forward DFS from StartEvent and from external_step). external_step is created when a step accepts a HumanResponseEvent (not only when it returns InputRequiredEvent). Unreachable steps raise WorkflowValidationError.

Rule 2 — No dead ends: Every step that produces events must have a path to an output event (reverse DFS from StopEvent / InputRequiredEvent). Steps that only return None are exempt. Terminal-event check ensures only StopEvent and InputRequiredEvent are allowed as terminal.

Opt-out: Workflow(..., skip_graph_checks={"reachability", "terminal_event", "dead_end"}) or @step(skip_graph_checks=["reachability"] | ["dead_end"]).

Caching: First successful validate() is cached; cache cleared when add_step() is used (via _step_functions_version).


packages/llama-index-workflows/src/workflows/decorators.py
```diff
@@ -76,6 +93,7 @@ def step(
     workflow: Type["Workflow"] | None = None,
     num_workers: int = 4,
     retry_policy: RetryPolicy | None = None,
+    skip_graph_checks: list[str] | None = None,
```
adrianlyjak: this should be a `list[Literal["reachability", "dead_end"]]` so values are discoverable

packages/llama-index-workflows/src/workflows/decorators.py
```diff
@@ -36,6 +36,8 @@
     from .workflow import Workflow
 from .retry_policy import RetryPolicy
 
+_VALID_STEP_GRAPH_CHECKS: set[str] = {"reachability", "dead_end"}
```
adrianlyjak: any reason these are different than the other set? This looks like odd duplication

packages/llama-index-workflows/src/workflows/representation/build.py
```diff
@@ -307,6 +307,16 @@ def _ensure_descriptor_node(
                 )
                 added_nodes.add(event_type.__name__)
 
+            # Add external_step node when a step accepts a HumanResponseEvent
```
adrianlyjak: hmm, this seems like it shouldn't be necessary? And the fixed id seems odd. Those are supposed to be unique. So we're sort of combining all external steps into one node, and it doesn't seem to actually be connected?

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -458,6 +478,123 @@ def run(
...
+
+        Input events: StartEvent, HumanResponseEvent (via external_step). Output events:
+        StopEvent, InputRequiredEvent. Steps returning None (e.g. HumanResponseEvent mutation
+        pattern) don't produce event nodes in the graph, so they are naturally allowed as
+        terminal steps without special-case handling.
+        """
+        from workflows.representation import get_workflow_representation
+        from workflows.representation.types import WorkflowEventNode, WorkflowStepNode
+
+        graph = get_workflow_representation(self)
```
adrianlyjak: Hmm, this `get_workflow_representation` function is doing some heavy/magical weight things that are more for development/debugging, and would like to avoid them happening in runtime scenarios. Things like dynamic type programming with inspect, looking up source code, and analyzing dependencies.

There's probably a small refactor or sub function that can can be called to focus more on just extracting the step graph from its already defined type info. Ideally this should operate on some sort of other data structure composed mostly of the `StepConfig` or `StepFunction` types.

packages/llama-index-workflows/tests/test_workflow.py
```diff
@@ -916,3 +916,310 @@ async def other_step(self, ctx: Context, ev: StartEvent) -> None:
...
+    """HumanResponseEvent step returning None (mutation pattern) is allowed."""
+
+    class HITLMutationWorkflow(Workflow):
+        @step
+        async def ask(self, ev: StartEvent) -> InputRequiredEvent:
+            return InputRequiredEvent()
+
+        @step
+        async def mutate(self, ev: HumanResponseEvent) -> None:
+            return None  # type: ignore[return-value]
```
adrianlyjak: this doesn't look like it needs an ignore return value?

packages/llama-index-workflows/tests/test_workflow.py
```diff
@@ -916,3 +916,310 @@ async def other_step(self, ctx: Context, ev: StartEvent) -> None:
...
+    class WorkflowLevelOptOut(Workflow):
+        @step
+        async def entry(self, ev: StartEvent) -> _GraphValidationProcessedEvent:
+            return _GraphValidationProcessedEvent()
+
+        @step
+        async def finish(self, ev: _GraphValidationProcessedEvent) -> StopEvent:
+            return StopEvent(result="done")
+
+        @step(skip_graph_checks=["dead_end"])
```
adrianlyjak: this seems unnecessary if the whole workflow opted out

vasudev13:
Hey @adrianlyjak 

I think the CI fails because an existing test (test_decorated_config) adds a step to DummyWorkflow as a side effect. With the new graph validation, that extra step gets flagged as unreachable when another test reuses DummyWorkflow in the same xdist worker. 

Let me know if I am wrong.

packages/llama-index-workflows/src/workflows/decorators.py
vasudev13: Agreed.

packages/llama-index-workflows/src/workflows/decorators.py
vasudev13: _VALID_STEP_GRAPH_CHECKS constant is gone. Validation now relies on typing.get_args(WorkflowGraphCheck)

packages/llama-index-workflows/tests/test_workflow.py
vasudev13: `test_graph_validation_human_response_event_mutation_allowed` now has a simple return `None`.

packages/llama-index-workflows/src/workflows/representation/build.py
vasudev13: The external_step node is no longer used by validation — graph checks now build directly from StepConfig. It still exists in build.py for visualization. Want me to clean it up there as part of this PR ?

packages/llama-index-workflows/src/workflows/workflow.py
vasudev13: refactored to build own adjacency list by iterating `self._get_steps()` and reading StepConfig.accepted_events / StepConfig.return_types directly — no reflection, no source parsing, just the type info already stored on the step decorator metadata.

vasudev13:
Thanks for the change @adrianlyjak

adrianlyjak:
@vasudev13 Thanks for this change! I'm going to merge it in, and follow up with some organization/type refactors before release

> I think the CI fails because an existing test (test_decorated_config) adds a step to DummyWorkflow as a side effect. 

yup, it was just test pollution. Not related to xdist, happens serially too

PR #445, opened 2026-03-23
Export workflow instrumentation event types from public API

enrico-stauss:
WorkflowStepOutputEvent, WorkflowRunOutputEvent, and SpanCancelledEvent are emitted through the LlamaIndex instrumentation dispatcher during workflow execution. Observability integrations such as openinference-instrumentation-llama-index use singledispatch to handle known event types and log a warning for any type they don't recognise:
- WARNING - Unhandled event of type WorkflowStepOutputEvent
- WARNING - Unhandled event of type WorkflowRunOutputEvent

The [root fix belongs in the openinference package](https://github.com/Arize-ai/openinference/pull/2908) (add no-op or meaningful handlers for these types). Exporting the classes from `workflows.__init__` gives the openinference maintainers a stable public import path:
- from workflows import WorkflowStepOutputEvent, WorkflowRunOutputEvent, SpanCancelledEvent

so they can register singledispatch handlers without importing from internal implementation modules.

### DIFF
--- a/packages/llama-index-workflows/src/workflows/__init__.py
+++ b/packages/llama-index-workflows/src/workflows/__init__.py
@@ -5,6 +5,11 @@ from pkgutil import extend_path
 
 from .context import Context
 from .decorators import step
+from .runtime.types.step_function import (
+    SpanCancelledEvent,
+    WorkflowRunOutputEvent,
+    WorkflowStepOutputEvent,
+)
 from .workflow import Workflow
 
 __path__ = extend_path(__path__, __name__)
@@ -12,6 +17,9 @@ __path__ = extend_path(__path__, __name__)
 
 __all__ = [
     "Context",
+    "SpanCancelledEvent",
     "Workflow",
+    "WorkflowRunOutputEvent",
+    "WorkflowStepOutputEvent",
     "step",
 ]


adrianlyjak:
@enrico-stauss I'm not sure it makes sense to make these events public. At the very least, we carefully curate what's available as a root module import, and these do not belong there (perhaps exported from a separate more-out-of-the-way, but still stable file)

Alternatively, at least on the llama-index-workflows side, we've significantly improved our llama-index-observability-opentelemetry backend, and it now has detailed event details like the arize openinference telemetry adds. You could consider switching to that (arize supports open telemetry collectors). Note however it may not instrument llama_index based abstractions like arize does.

enrico-stauss:
I'll give it a try and come back to you, thanks.

adrianlyjak:
@enrico-stauss see also https://github.com/Arize-ai/openinference/pull/2908#issuecomment-4120446348. The coupling here seems hard to maintain well

enrico-stauss:
Hi @adrianlyjak 
I migrated my application to `llama-index-observability-otel` but my finding is, that if I do so, I loose all of the commodities like token counts, LLM completions and so on. I would have to write a custom span exporter, correct?

I might do that at some point, but for now I decided to defer the migration and just silence the warning. I get that the situation where a fix spans 2 or 3 repos is not maintainable, but the current alternative is not yet mature (unless I missed something). You may close this PR if you want to.

Cheers
Enrico

adrianlyjak:
@enrico-stauss yeah, aware that llama-index internals need some rework for parity with arize 😢 . In general I think the correct approach is to not log warnings based on an external dependency. Unfortunately I don't think that's likely to happen, they recently [changed contribution policy](https://github.com/Arize-ai/openinference/blob/d563e428983808b9151d2b7c868d3778c33441a8/CONTRIBUTING#L5). I don't expect updates to their adapter in the near future

Issue #454, opened 2026-03-31
[RFC] Add retryable_exceptions filter to retry policies


vasudev13:
### Motivation

Both built-in policies retry on *any* exception. This means a `ValueError` from a malformed response or a `KeyError` in parsing logic burns through all attempts even though retrying will never help — only transient failures like network timeouts or rate limits warrant a retry.

### Proposed Change

Add an optional `retryable_exceptions` parameter to both concrete policy classes. Default `None` preserves existing behavior, no breaking changes. 

Since `workflows-py` is LLM-agnostic, the parameter accepts any exception type(s) the caller needs. Example - 

**OpenAI**
```python
from openai import RateLimitError, APITimeoutError, APIConnectionError

@step(retry_policy=ExponentialBackoffRetryPolicy(
    maximum_attempts=5,
    retryable_exceptions=(RateLimitError, APITimeoutError, APIConnectionError),
))
async def call_openai(self, ev: QueryEvent) -> ResultEvent: ...
```

**Anthropic**
```python
from anthropic import RateLimitError, APIConnectionError

@step(retry_policy=ConstantDelayRetryPolicy(
    delay=10,
    maximum_attempts=3,
    retryable_exceptions=(RateLimitError, APIConnectionError),
))
async def call_anthropic(self, ev: QueryEvent) -> ResultEvent: ...
```

**Gemini**
```python
from google.api_core.exceptions import ResourceExhausted, ServiceUnavailable

@step(retry_policy=ExponentialBackoffRetryPolicy(
    maximum_attempts=5,
    retryable_exceptions=(ResourceExhausted, ServiceUnavailable),
))
async def call_gemini(self, ev: QueryEvent) -> ResultEvent: ...
```

Implementation is ~3 lines per policy — early-return `None` if `not isinstance(error, self.retryable_exceptions)`.

### Questions

1. Is this in scope for the built-in policies, or is the intent that users subclass `RetryPolicy` for this level of control?
2. Happy to open a PR if the direction looks good — I contributed the `ExponentialBackoffRetryPolicy` previously so I know this surface area well.

adrianlyjak:
Thanks for bringing this up @vasudev13. Definitely agree that the retries are pretty lacking right now.

One thing that gives me pause with this design is the detection of whether an exception is retryable ends up embedded into the retry policy. If we continue extending, or a user creates their own retry policy, each retry policy needs to re-implement this logic.

[...]

PR #482, opened 2026-04-06
feat: Retry Policy

vasudev13:
## Summary

Implements composable retry policy primitives inspired by tenacity's API, as discussed in #454. This decomposes retry behavior into three orthogonal concerns:

- **Retry conditions**: `retry_if_exception_type`, `retry_if_not_exception_type`, `retry_if_exception_message` — decide *whether* to retry
- **Wait strategies**: `wait_fixed`, `wait_exponential`, `wait_random`, `wait_exponential_jitter`, `wait_chain` — decide *how long* to wait
- **Stop conditions**: `stop_after_attempt`, `stop_after_delay` — decide *when* to give up

These compose into a unified `RetryPolicy` class:

```python
@step(retry_policy=RetryPolicy(
    retry=retry_if_exception_type(RateLimitError, APITimeoutError),
    wait=wait_exponential(initial=1, multiplier=2, max=60),
    stop=stop_after_attempt(5),
))
async def call_api(self, ev: StartEvent) -> StopEvent: ...

### CODE REVIEW

**Comment by adrianlyjak** [comment]
I think this direction seems like a good one! Only thing is that the compatibility with the class and the renamed protocol might be messy. Need to think through ramifications on library import usage, or if there’s any retry policy protocol implementations in the wild

**Comment by vasudev13** [comment]
Good call — pushed a fix. I reverted the RetryPolicy → RetryPolicyProto rename so RetryPolicy stays the structural Protocol (fully backward compatible for anyone annotating against it or implementing their own), and renamed the new composable class to ComposableRetryPolicy. Usage is now:


```
@step(retry_policy=ComposableRetryPolicy(
    retry=retry_if_exception_type(RateLimitError),
    wait=wait_exponential(initial=1, multiplier=2, max=60),
    stop=stop_after_attempt(5),
))

```

adrianlyjak:
@vasudev13 thanks. I'm going to push some more tweaks here to the structure. I have a reasonable plan to make the composable retry policy the only concrete class, but keep it private, and `RetryPolicy(...)` the constructor. This should be the simplest user interface. Also looks like there may be some divergence from tenacity methods. I'll add some tests to validate that we conform to the same interfaces

adrianlyjak:
> and RetryPolicy(...) the constructor

That ended up being too hacky. Rather, introduced a `retry_policy(...)` constructor to match the functional appearance of the composable retry constructor functions

Also: 
- Added test to ensure tenacity conformance, 
- added combinators
- as well as some missing tenacity parity constructors
- added docs

vasudev13:
Thanks so much @adrianlyjak for your feedback and work on this PR!!

PR #483, opened 2026-04-06
docs(examples): give every example a folder and README

adrianlyjak:


adrianlyjak:
eh, this is kind of worse

PR #554, opened 2026-04-16
feat(workflows): add @catch_error handler + Context.retry_info()

### LINKED ISSUE #12: [Feature Request]: Better way to handle exception on workflows
### Feature Description

Implement an exception handler that includes all step context, allowing for user customization.

### Reason

Using workflows we can use retry_policy, but when max retries on a step exceed I can't handle this exceptions and send to another step, for example.

My workaround for now is using another workflow to handle exceptions in a nested workflow, but this workflow exists just for that. Another drawback is that when retry happens, I lost data in the context during that try that raises exception.

The current retry_policy lacks a direct mechanism to handle max-retry exceptions inline. I'm looking for a solution that avoids an extra nested workflow and preserves context data when rerouting after a failed retry.

### Value of Feature

exception handler for workflows

adrianlyjak:
## Summary

- `@catch_error` decorator registers a terminal handler that fires when any step exhausts its retry policy; it may return a `StopEvent` to recover or raise to fail the workflow.
- `Context.retry_info()` surfaces the current attempt number, elapsed time, and the most recent `FailureInfo` from inside step bodies. Survives workflow serialize/resume.
- `StepFailedEvent` carries step name, input-event type + JSON dump (with a fallback marker for non-serializable events), exception type/message/traceback, attempt count, elapsed seconds.
- Workflow construction rejects >1 `@catch_error` handlers or non-`StopEvent` return types.
- Control loop integration: on retry exhaustion with a handler registered, routes `StepFailedEvent` directly to the handler step via `CommandQueueEvent(step_name=...)`. Recursion-guarded — handler failures fall through to the existing `WorkflowFailedEvent` path. Timeouts and cancellations unchanged.

Closes run-llama/llama-agents#12.


packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -86,6 +90,68 @@
 from workflows.workflow import Workflow
 
 
+def _dump_event_safely(event: Event) -> dict[str, Any]:
```
adrianlyjak, line 93: this is some garbage. we serialize events all over the place. Now is not the time nor place to be paranoid

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -86,6 +90,68 @@
...
+    step_name: str,
+    triggering_event: Event,
+    exception_type: str,
+    exception_message: str,
+    exc_traceback: str,
+    attempts: int,
+    elapsed_seconds: float,
+) -> StepFailedEvent:
+    event_cls = type(triggering_event)
+    input_event_type = f"{event_cls.__module__}.{event_cls.__qualname__}"
```
adrianlyjak, line 127: hmm, look around for utils that assist in this. EventEnvelope/JsonSerializer (EventEnvelope I think is not in a reachable package though)

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -86,6 +90,68 @@
...
+        input_event=_dump_event_safely(triggering_event),
+        exception_type=exception_type,
+        exception_message=exception_message,
+        traceback=exc_traceback,
+        attempt=attempts,
+        elapsed_seconds=elapsed_seconds,
+    )
+
+
+def _build_failure_info(exception: BaseException, failed_at: float) -> FailureInfo:
```
adrianlyjak, line 140: this needs a better home

packages/llama-index-workflows/src/workflows/__init__.py
```diff
@@ -4,14 +4,20 @@
...
+from .errors import FailureInfo, RetryInfo
+from .events import StepFailedEvent
 from .workflow import Workflow
 
 __path__ = extend_path(__path__, __name__)
 
 
 __all__ = [
     "Context",
+    "FailureInfo",
```
adrianlyjak, line 17: uh, let's keep these exported from just events

packages/llama-index-workflows/src/workflows/errors.py
```diff
@@ -42,3 +46,39 @@ class ContextStateError(Exception):
     - ExternalContext: During run, for handler/external code
     - InternalContext: During run, for step execution
     """
+
+
+@dataclass(frozen=True)
+class FailureInfo:
```
adrianlyjak, line 52: these others are actual errors. Also seems like potentially the wrong location/organization

packages/llama-index-workflows/src/workflows/events.py
```diff
@@ -291,6 +291,36 @@ class WorkflowFailedEvent(StopEvent):
...
+        exception_type: Fully qualified type name of the raised exception.
+        exception_message: `str(exception)`.
+        traceback: Joined `traceback.format_exception(...)` output.
+        attempt: Total number of attempts made before giving up.
+        elapsed_seconds: Seconds from first attempt to final failure.
+    """
+
+    step_name: str
+    input_event_type: str
+    input_event: dict[str, Any]
```
adrianlyjak, line 316: should look into serialization utils so this can remain an Event type, this is ugly

packages/llama-index-workflows/src/workflows/events.py
```diff
@@ -291,6 +291,36 @@ class WorkflowFailedEvent(StopEvent):
...
+    exception_type: str
+    exception_message: str
```
adrianlyjak, line 318: exception should be a single sub field, more representing a serialized exception.

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -342,6 +344,38 @@ def _ensure_stop_event_class(self) -> type[RunResultT]:
...
+            if step_func._step_config.role == "catch_error":
+                catch_error_steps.append(name)
+
+        if len(catch_error_steps) > 1:
+            names = ", ".join(sorted(catch_error_steps))
+            raise WorkflowValidationError(
+                f"Only one @catch_error handler is allowed per workflow, found {len(catch_error_steps)}: {names}"
+            )
+
+        if len(catch_error_steps) == 1:
```
adrianlyjak, line 365: this is dumb. Remove it. Make this part of the _graph validation_, and make sure to test it there

packages/llama-index-workflows/src/workflows/context/context.py
```diff
@@ -574,6 +575,29 @@ async def my_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
             event_type, waiter_event, waiter_id, requirements, timeout
         )
 
+    def retry_info(self) -> RetryInfo:
+        """Return a snapshot of the currently-executing step's retry state.
+
+        Returns:
+            RetryInfo: attempt number (1-based), seconds since the first attempt,
```
adrianlyjak: hmm, re-evaluate what is most internally consistent,  whether this should be 0 or 1. Should prefer 0 (more standard in python)

packages/llama-index-workflows/src/workflows/runtime/types/serialization_helpers.py
```diff
@@ -13,46 +13,20 @@
...
+)
 
-SerializableOptionalEvent = Annotated[
-    Event | None,
-    PlainSerializer(_serialize_optional_event, return_type=Any),
-    PlainValidator(_deserialize_optional_event),
+__all__ = [
+    "SerializableEvent",
+    "SerializableOptionalEvent",
+    "SerializableException",
```
adrianlyjak: hmm, we should mark/comment these are for back compat and that they will be removed. Don't add new types here

packages/llama-index-workflows/src/workflows/runtime/types/step_function.py
```diff
@@ -108,6 +109,10 @@ def __call__(
         step_name: str,
         event: Event,
         workflow: Workflow,
+        retry_number: int = 0,
```
adrianlyjak: I'm finding these new params very noisy. Should we wrap them in a type? What's on the StepWorkerState? Could it go on there?

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -742,44 +752,67 @@ def _process_step_result_tick(
...
-                            attempts=total_attempts,
-                            elapsed_seconds=elapsed,
+
+                catch_error_name = state.config.catch_error_step_name
+                should_route = (
+                    catch_error_name is not None and tick.step_name != catch_error_name
+                )
+                if should_route:
+                    # Route to the catch-error handler. Keep workflow running so
+                    # the handler can produce either a StopEvent or a new failure.
```
adrianlyjak, line 844: what happens if this in turn fails? Will this infinite loop?

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -640,6 +646,17 @@ def _validate(
                         steps_accepting_stop_event.append(name)
                         break
 
+            if step_config.role == "catch_error":
+                catch_error_steps.append(name)
+                for rt in step_config.return_types:
+                    if rt is type(None):
+                        continue
+                    if not (isinstance(rt, type) and issubclass(rt, StopEvent)):
```
adrianlyjak: uh, this was not the intent. Should be able to spawn its own graph that must be terminal. This should in general be managed by the graph validator, not here inline anyways

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -672,7 +700,10 @@ def _validate(
         unconsumed_events = {
             x
             for x in unconsumed_events
-            if not issubclass(x, (InputRequiredEvent, HumanResponseEvent, StopEvent))
+            if not issubclass(
+                x,
+                (InputRequiredEvent, HumanResponseEvent, StopEvent, StepFailedEvent),
```
adrianlyjak, line 756: this doesn't need to go here. This is nonsense, should not be producing `StepFailedEvent`s either way. Same really for `HumanResponseEvent`, that doesn't belong here actually

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -650,13 +680,42 @@ def _validate(
...
                 produced_events.add(event_type)
 
-        # Raise error if any steps incorrectly accept StopEvent
         if steps_accepting_stop_event:
             step_names = "', '".join(steps_accepting_stop_event)
             plural = "" if len(steps_accepting_stop_event) == 1 else "s"
             msg = f"Step{plural} '{step_names}' cannot accept StopEvent. StopEvent signals the end of the workflow. Use a different Event type instead."
             raise WorkflowValidationError(msg)
 
+        from .representation.validate import validate_catch_error_handlers
```
adrianlyjak: yuck, why's this in here

packages/llama-index-workflows/src/workflows/workflow.py
```diff
@@ -650,13 +680,42 @@ def _validate(
...
+        handler_errors = validate_catch_error_handlers(
+            catch_error_handlers, all_step_names
+        )
+        if handler_errors:
+            raise WorkflowValidationError("\n".join(handler_errors))
+
+        handler_step_names = {h.step_name for h in catch_error_handlers}
+        handler_for_step: dict[str, str] = {}
+        for handler in catch_error_handlers:
+            if handler.for_steps is None:
```
logan-markewich: why is for_steps optional? Shouldn't it always have specific step names or all step names?

logan-markewich (approved):
looks legit to me honestly

packages/llama-index-workflows/src/workflows/workflow.py
adrianlyjak: `None` means all "other" step names here (`wildcard` below). Computed after the fact as the remaining steps

packages/llama-index-workflows/src/workflows/runtime/control_loop.py
```diff
@@ -787,44 +808,83 @@ def _process_step_result_tick(
             else:
                 delay = None
             if delay is not None:
+                exception_info = ExceptionInfo.from_exception(result.exception)
```
adrianlyjak: This is dumb. 2 serializable exception types. One for holding a stack trace, the other not, but the stack trace holding one is initialized from the other that no longer has a stack. This needs to be redesigned

PR #574, opened 2026-04-26
Add fan-out / fan-in syntax proposal

adrianlyjak:
Vision doc for simplifying map-reduce patterns: list/AsyncIterator
returns, parameter-typed fan-in via list[E], and a Collect annotation
for non-default scope/cardinality/provenance. Implementation deferred;
this is for team review of the API surface.

### DIFF
--- a/architecture-docs/proposals/fan-out-fan-in.md
+++ b/architecture-docs/proposals/fan-out-fan-in.md
@@ -161,13 +161,26 @@ class Collect:
 
 ## AsyncIterator semantics
 
-Async-iterator fan-out is the elegant default — it lets the type graph stay honest while supporting dynamic cardinality. It introduces one wrinkle: failure mid-batch.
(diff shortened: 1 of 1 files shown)


architecture-docs/proposals/fan-out-fan-in.md
```diff
@@ -0,0 +1,239 @@
...
+
+```python
+@step
+async def join(events: list[Done]) -> StopEvent:
+    return StopEvent(results=events)
+```
+
+That's the whole common case. No decorator argument, no buffer id, no counter in `ctx.store`.
+
+## How it reads in practice
```
logan-markewich, line 75: Honestly im a fan of the syntax.  But I feel every time weve attempted this, tracking events under the hood was really hard, especially for nested fan ins/outs. My mind keeps wondering to the implementation :p

logan-markewich: Maybe workflows 2.0 makes this easier, last time we tried it was with the original code

adrianlyjak: I think some things with the control loop structure will make it a bit easier, but definitely not going to be trivial

PR #586, opened 2026-04-27
llamactl: deployments get/list/logs

adrianlyjak:
## Summary

- `deployments get <name>` shows a single deployment (no auto-launching the TUI; `deployments logs --follow` or `deployments edit` get you the live view).
- `deployments get` (no args) prints a kubectl-style table of all deployments. `list` stays as a hidden alias.
- New `deployments logs <name>`: prints recent logs and exits by default; `--follow` to stream. Also `--json` (jsonl), `--tail`, `--since-seconds`, `--include-init-containers`.
- `-o text|json|yaml|wide` on `get` / `list` / `logs`. Text excludes `wide=True` columns; `wide` interleaves them into their natural positions.
- `--project <id>` overrides the active profile's project for one command.
- 404 from the control plane renders as `Deployment <id> not found in project <project>` instead of the verbose URL + body dump. Other status codes keep the verbose message.
- Per-command display models (`DeploymentDisplay` / `DeploymentSpec` / `DeploymentStatus`) layered on top of the column framework. Secrets surface as masked `{name: "********"}` dicts; `personal_access_token` shows as `********` when set, omitted when not.

Stacked on:
- adrian/llamactl-cli-infra (#585) — the column framework / render helpers / log_format / `--project` plumbing.
- adrian/server-logs-follow-flag (#584) — needed at runtime for `deployments logs` (no-`--follow`) to bound the request server-side.

### DIFF
--- a/.changeset/llamactl-deployments-get-logs.md
+++ b/.changeset/llamactl-deployments-get-logs.md
@@ -2,4 +2,4 @@
 "llamactl": minor
 ---
 
-`deployments get` now does double duty: with a name it shows that deployment (no auto-launching the TUI); without one it prints a kubectl-style table of all deployments. Adds `-o text|json|yaml|wide`, `--project <id>` to override the active profile, friendlier 404 messages, and a new `deployments logs` command (`--follow`/`--json`/`--tail`/`--since-seconds`/`--include-init-containers`). The old `deployments list` is kept as a hidden alias.
+`deployments get` now shows one deployment with a name and lists all of them without; adds `-o text|json|yaml|wide`, `--project`, and a new `deployments logs` command.
--- a/.changeset/llamactl-deployments-get-logs.md
+++ b/.changeset/llamactl-deployments-get-logs.md
@@ -1,0 +1,5 @@
+---
+"llamactl": minor
+---
+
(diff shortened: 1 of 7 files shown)


packages/llamactl/src/llama_agents/cli/display.py
```diff
@@ -162,3 +168,107 @@ def render_columns(
...
+    model_config = ConfigDict(extra="forbid")
+
+    display_name: str
+    repo_url: Annotated[str, Column("REPO", format=gh_short)]
+    deployment_file_path: str
+    git_ref: Annotated[str | None, Column("GIT_REF", default="-")] = None
+    appserver_version: Annotated[
+        str | None, Column("APPSERVER", default="-", wide=True)
+    ] = None
+    suspended: Annotated[bool, Column("SUSPENDED", wide=True)] = False
```
adrianlyjak: this should not be in wide (is already represented in the status)

.changeset/llamactl-deployments-get-logs.md
```diff
@@ -0,0 +1,5 @@
+---
+"llamactl": minor
+---
+
+`deployments get` now does double duty: with a name it shows that deployment (no auto-launching the TUI); without one it prints a kubectl-style table of all deployments. Adds `-o text|json|yaml|wide`, `--project <id>` to override the active profile, friendlier 404 messages, and a new `deployments logs` command (`--follow`/`--json`/`--tail`/`--since-seconds`/`--include-init-containers`). The old `deployments list` is kept as a hidden alias.
```
adrianlyjak: this is too verbose

PR #604, opened 2026-04-30
llamactl: deployments apply -f

adrianlyjak:
Adds file-driven deployment apply and delete.

`deployments apply -f` uses the same shape emitted by `deployments get -o yaml` and `deployments template`, so the workflow is render YAML, edit it, apply it back. The parser resolves `${VAR}` references locally, ignores existing `********` mask sentinels on round-trip, and treats `null` secret values as deletes.

Apply is a client-side upsert. The CLI checks whether the deployment exists, validates external repo URLs before save, then creates or updates. Push-mode is part of the same path: `repo_url: ""` means push the current working tree to the internal repo, with the ordering chosen so the server can resolve `git_ref` against the right commits.

`deployments delete -f` reads the top-level `name` from the same YAML shape and deletes that deployment. The file flag is mutually exclusive with the positional deployment arg.

### DIFF
--- a/packages/llamactl/src/llama_agents/cli/apply_yaml.py
+++ b/packages/llamactl/src/llama_agents/cli/apply_yaml.py
@@ -1,0 +1,142 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+"""YAML parsing for ``llamactl deployments apply -f``.
+
+Pure parsing + validation — no network, no client calls.  Takes YAML text,
+resolves ``${VAR}`` environment variables, strips ``********`` mask sentinels,
+and validates against :class:`DeploymentDisplay`.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from typing import Any
+
+import yaml
+from llama_agents.cli.display import DeploymentDisplay, DeploymentSpec
+from pydantic import ValidationError
+
+_ENV_VAR_RE = re.compile(r"\$\{([^}]+)\}")
+
+
+class ApplyYamlError(Exception):
+    """Base error for YAML apply parsing/validation failures."""
+
+
+class UnresolvedEnvVarsError(ApplyYamlError):
+    """Raised when ``${VAR}`` references cannot be resolved."""
+
+    def __init__(self, unresolved: list[str]) -> None:
+        self.unresolved = unresolved
+        super().__init__(
+            f"unresolved environment variables: {', '.join(sorted(unresolved))}"
+        )
+
+
+_ENV_STRING_SPEC_FIELDS = (
+    "repo_url",
+    "deployment_file_path",
+    "git_ref",
+    "appserver_version",
+    "personal_access_token",
+)
+
+
+def _resolve_string(text: str, unresolved: list[str]) -> str:
+    def _replacer(match: re.Match[str]) -> str:
+        var = match.group(1)
+        env_val = os.environ.get(var)
+        if env_val is None:
+            unresolved.append(var)
+            return match.group(0)  # leave ${VAR} as-is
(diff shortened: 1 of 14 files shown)


packages/llamactl/src/llama_agents/cli/yaml_format.py
```diff
@@ -0,0 +1,296 @@
...
+
+    return out
+
+
+# ---------------------------------------------------------------------------
+# Input alias normalization
+# ---------------------------------------------------------------------------
+
+
+def _normalize_aliases(data: dict[str, Any]) -> dict[str, Any]:
```
adrianlyjak: this is dumb. Let's have a single standard name

packages/llamactl/src/llama_agents/cli/yaml_format.py
```diff
@@ -0,0 +1,296 @@
...
+
+    return _ENV_VAR_RE.sub(_replacer, text)
+
+
+# ---------------------------------------------------------------------------
+# Mask stripping
+# ---------------------------------------------------------------------------
+
+
+def _strip_masks(data: dict[str, Any]) -> dict[str, Any]:
```
adrianlyjak: isn't this already happening elsewhere?

packages/llamactl/src/llama_agents/cli/apply_yaml.py
```diff
@@ -0,0 +1,296 @@
...
+
+    return DeploymentCreate(
+        id=display.name,
+        display_name=display_name,
+        repo_url=repo_url,
+        **spec_data,
+    )
+
+
+def apply_payload_to_update(display: DeploymentDisplay) -> DeploymentUpdate:
```
adrianlyjak: why is this file called yaml_format? I think we need to refactor our organizational structures. This is getting to be a convoluted mess

packages/llamactl/src/llama_agents/cli/commands/deployment.py
```diff
@@ -377,6 +406,253 @@ def delete_deployment(
...
+            )
+        )
+        click.echo(verdict)
+        return
+
+    validate_authenticated_profile(interactive=False)
+    client = get_project_client(project_id_override=project)
+
+    # Deferred: llamactl startup budget avoids importing httpx at module level.
+    import httpx
```
adrianlyjak, line 456: this should go to the top of the method. Update AGENTS.md accordingly

packages/llamactl/src/llama_agents/cli/commands/deployment.py
```diff
@@ -91,6 +101,23 @@ def friendly_http_error(
     return msg
 
 
+def _dry_run_spec_dump(spec: DeploymentSpec) -> dict[str, Any]:
```
adrianlyjak: quit this weird untyped after serialization stuff.  Push this into the model

something like
```
spec.as_redacted().model_dump("json", exclude_unset=True)
```

packages/llamactl/src/llama_agents/cli/commands/deployment.py
```diff
@@ -377,6 +425,219 @@ def delete_deployment(
...
+    try:
+        _apply_push(client, deployment_id, git_ref)
+    except click.ClickException as exc:
+        click.echo(str(exc.message), err=True)
+        raise click.ClickException(
+            "re-run `llamactl deployments apply -f <file>` to retry the push"
+        ) from exc
+
+
+async def _apply_deployment_from_yaml(client: Any, display: DeploymentDisplay) -> None:
```
adrianlyjak: this should not be `Any`, this is lazy. We can use TYPE_CHECKING to keep the type deferred

PR #608, opened 2026-04-30
Add opt-in workflow pressure diagnostics

adrianlyjak:
Workflow pressure is hard to debug from process metrics alone. This adds opt-in instrumentation in `llama-agents-server` that records workflow step intervals next to event-loop lag and RSS threshold events, without adding routes or starting background sampling by default.

- Adds `PressureDiagnosticsConfig` and a process-local recorder, then decorates default or custom runtimes only when diagnostics are enabled.
- Tracks step start/finish intervals from `StepStateChanged`, with cleanup on close/cancel and bounded snapshot/recent-event APIs for a later debugger route.
- Samples RSS and event-loop lag with rate-limited structured logs, optional loop stack capture, and overlap-based interval payloads.
- Documents enablement and adds recorder/runtime tests.

Tests:
- `uv run dev -p llama-agents-server -- tests/server/test_pressure_diagnostics.py -q`
- `uv run dev -p server`
- `uv run ty check packages/llama-agents-server/src packages/llama-agents-server/tests/server/test_pressure_diagnostics.py`
- `uv run ruff check packages/llama-agents-server/src/llama_agents/server packages/llama-agents-server/tests/server/test_pressure_diagnostics.py`
- `uv run pre-commit run -a` fails in existing repo-wide type hooks on unresolved optional imports outside this server change (`bedrock_agentcore`, `cryptography`, `llamactl`/`textual`).

### DIFF
--- a/packages/llama-agents-server/src/llama_agents/server/_diagnostics/__init__.py
+++ b/packages/llama-agents-server/src/llama_agents/server/_diagnostics/__init__.py
@@ -1,0 +1,16 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+
+from __future__ import annotations
+
+from .pressure import (
+    PressureDiagnosticsConfig,
+    PressureDiagnosticsDecorator,
+    PressureDiagnosticsRecorder,
+)
+
+__all__ = [
+    "PressureDiagnosticsConfig",
+    "PressureDiagnosticsDecorator",
+    "PressureDiagnosticsRecorder",
+]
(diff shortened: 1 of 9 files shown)


packages/llama-agents-server/src/llama_agents/server/_diagnostics/__init__.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-License-Identifier: MIT
```
adrianlyjak, line 1: this file doesn't look useful and is just noisy extra code for import/export

PR #656, opened 2026-06-01
support subclass-aware event routing

vanrajsinh650:
## Problem

Workflow event validation and routing used exact event type checks in several paths.

This caused workflows to reject or ignore subclass events. For example, if one step returns `ChildEvent` and the next step accepts `ParentEvent`, the workflow should treat `ChildEvent` as valid because it subclasses `ParentEvent`.

```python
class ParentEvent(Event):
    value: str

class ChildEvent(ParentEvent):
    pass
```
## Before this change, the workflow failed during validation with:

WorkflowValidationError: The following events are consumed but never produced: ParentEvent

This happened even though ChildEvent was produced and ChildEvent is a subclass of ParentEvent.

## Solution

Use subclass-aware event matching across workflow validation and runtime routing.

## Changes include:

Use issubclass(produced_type, consumed_type) for graph connectivity validation.
Use isinstance(event, accepted_type) for runtime step routing.
Use isinstance(event, waiting_for_event) for waiter resolution.
Use isinstance(message, accepted_type) for targeted step dispatch validation.
Keep exact-type routing behavior working, since issubclass(X, X) and isinstance(x, X) still support exact matches.
Tests

## Added regression coverage for:

subclass event routing to a step accepting the parent event
runtime routing with validation disabled
targeted step validation
graph edge construction
waiter resolution with subclass events
exact-type event routing continuing to work

## After the fix:

uv run dev -p workflows -- -k subclass
8 passed

Full workflow test suite:

uv run dev -p workflows
610 passed, 94 warnings

Code quality:

uv run ruff check packages/llama-index-workflows
All checks passed!

uv run ruff format --check packages/llama-index-workflows
88 files already formatted

## Validation
uv run dev -p workflows -- -k subclass  (8 passed)
uv run dev -p workflows (610 passed, 94 warnings)
uv run ruff check packages/llama-index-workflows (passed)
uv run ruff format --check packages/llama-index-workflows (passed)

### DIFF
--- a/.changeset/brave-bees-sink.md
+++ b/.changeset/brave-bees-sink.md
@@ -1,5 +1,5 @@
 ---
-"llama-index-workflows": major
+"llama-index-workflows": minor
 ---
 
-Support subclass-aware workflow event routing and validation
+Support opt-in subclass event routing via the step decorator parameter `accept_event_subclasses`
--- a/.changeset/brave-bees-sink.md
+++ b/.changeset/brave-bees-sink.md
@@ -1,0 +1,5 @@
+---
+"llama-index-workflows": major
+---
+
+Support subclass-aware workflow event routing and validation
--- a/packages/llama-agents-server/tests/server/test_server_endpoints.py
+++ b/packages/llama-agents-server/tests/server/test_server_endpoints.py
@@ -31,7 +31,7 @@ from workflows import Context, step
 from workflows.context.context_types import SerializedContext
 from workflows.context.serializers import JsonSerializer
 from workflows.context.state_store import DictState, InMemoryStateStore
-from workflows.events import StartEvent, StopEvent
+from workflows.events import Event, StartEvent, StopEvent
 from workflows.workflow import Workflow
 
 
@@ -1420,3 +1420,65 @@ async def test_instrument_tags_contains_handler_id_in_server_context() -> None:
             assert data["status"] == "completed"
             assert seen_handler_id["handler_id"] is not None
             assert seen_handler_id["handler_id"] == handler_id
+
+
+# --- Subclass-aware event routing (accept_event_subclasses) over HTTP ---
+
+
+class SubclassParentEvent(Event):
+    payload: str
+
+
+class SubclassChildEvent(SubclassParentEvent):
+    pass
+
+
+class SubclassRoutingWorkflow(Workflow):
+    """A producer emits a subclass; a consumer accepts the parent with opt-in."""
+
+    @step
+    async def start(self, ev: StartEvent) -> SubclassChildEvent:
+        return SubclassChildEvent(payload="from-start")
+
+    @step(accept_event_subclasses=True)
+    async def consume(self, ev: SubclassParentEvent) -> StopEvent:
+        return StopEvent(result=f"consumed {type(ev).__name__}: {ev.payload}")
+
+
+@pytest_asyncio.fixture
+async def subclass_client() -> AsyncGenerator:
+    server = WorkflowServer(workflow_store=MemoryWorkflowStore(), idle_timeout=0.01)
+    server.add_workflow("subclass", SubclassRoutingWorkflow())
+    async with server.contextmanager():
+        transport = ASGITransport(app=server.app)
+        async with AsyncClient(transport=transport, base_url="http://test") as client:
+            yield client
+
+
+@pytest.mark.asyncio
+async def test_subclass_routing_runs_over_http(subclass_client: AsyncClient) -> None:
+    response = await subclass_client.post("/workflows/subclass/run", json={})
+    assert response.status_code == 200
+    data = response.json()
+    # The child event must have been routed to the parent-accepting opted-in step.
+    assert (
+        data["result"]["value"]["result"] == "consumed SubclassChildEvent: from-start"
+    )
+
+
+@pytest.mark.asyncio
+async def test_subclass_routing_representation_over_http(
+    subclass_client: AsyncClient,
+) -> None:
+    response = await subclass_client.get("/workflows/subclass/representation")
+    assert response.status_code == 200
+    graph = response.json()["graph"]
+    edges = {(e["source"], e["target"]) for e in graph["edges"]}
+    # Both the declared parent edge and the subclass fan-out edge are present.
+    assert ("SubclassParentEvent", "consume") in edges
+    assert ("SubclassChildEvent", "consume") in edges
+
+    child = next(n for n in graph["nodes"] if n["id"] == "SubclassChildEvent")
+    # The inheritance chain is serialized so the UI can classify the node.
+    assert "SubclassParentEvent" in child["event_types"]
+    assert child["event_schema"] is not None
--- a/packages/llama-index-workflows/src/workflows/_event_matching.py
+++ b/packages/llama-index-workflows/src/workflows/_event_matching.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+
+"""Centralized event-type matching.
+
+A workflow step declares the event types it accepts. Whether a given event
+routes to that step is decided in one of two modes:
+
+* **exact** (the default): the event matches only when its concrete type is
+  exactly one of the declared accepted types.
+* **subclass-aware** (``accept_event_subclasses=True``): the event also matches
+  when its type is a subclass of a declared accepted type.
+
+Every place that answers "does this event/type belong to this step?" — graph
+construction, static validation, and runtime routing — funnels through the
+helpers below so the rule stays identical across all of them.
+
+Step signatures may also carry annotations that are not classes at all
+(``dict``, ``list[str]``, ``typing.Any``, ...). ``issubclass`` raises
+``TypeError`` on those, so :func:`is_subclass` is used as a safe replacement
+that treats non-classes as "not a subclass" rather than letting them blow up.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+from typing import Any
+
+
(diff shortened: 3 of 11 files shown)


vanrajsinh650:
@adrianlyjak hey can you plz look at my pr

vanrajsinh650:
@AstraBert  hey can you plz look at my pr

adrianlyjak:
@vanrajsinh650 as implemented, this is a breaking change. 
1. Existing workflows that happen to have hierarchy will change behavior, so this cannot be merged as is. (We could consider some opt-in mechanisms to make it non-breaking, and consider making it the default in a future major version). 
2. visualization and workflow representation would need to be updated to understand this hierarchy
3. there's some UI layers in run-llama/llama-ui that may need tuning to understand hierarchy routing better

I am a little nervous about the additional complexity this introduces. Although, I do understand that it might allow for some more powerful patterns with hierarchy.  Can you share a little more about the end use case that you are trying to pattern with this change and how this is a poor developer experience currently for this end use case?

vanrajsinh650:
> @vanrajsinh650 as implemented, this is a breaking change. 
> 1. Existing workflows that happen to have hierarchy will change behavior, so this cannot be merged as is. (We could consider some opt-in mechanisms to make it non-breaking, and consider making it the default in a future major version). 
> 2. visualization and workflow representation would need to be updated to understand this hierarchy
> 3. there's some UI layers in run-llama/llama-ui that may need tuning to understand hierarchy routing better
> 
> I am a little nervous about the additional complexity this introduces. Although, I do understand that it might allow for some more powerful patterns with hierarchy.  Can you share a little more about the end use case that you are trying to pattern with this change and how this is a poor developer experience currently for this end use case?

Thanks for the detailed review that makes sense. I agree this should not change default routing behavior if existing workflows may depend on exact event matching.

The use case I was trying to support is workflows that model related event variants under a shared base event. For example:

DocumentEvent as a base event
InvoiceEvent, ReceiptEvent, and ContractEvent as subclasses
one shared step accepts DocumentEvent for common handling/logging/normalization more specific steps can still accept concrete event subclasses

Without subclass-aware routing, users need to either list every concrete event type explicitly or duplicate shared handling logic for each event variant. That can become hard to maintain as the workflow grows.

I think the breaking-change concern is valid. A safer direction could be making subclass-aware routing opt-in while preserving exact matching as the default.

Would you prefer this as a workflow-level option, a step-level option, or should this move to a design discussion first? I’m happy to adjust the PR based on the API shape you think fits best.

adrianlyjak:
Thanks, that use case makes sense.

One possible workaround today is a named union alias for the concrete variants:

```python
DocumentEvents = InvoiceEvent | ReceiptEvent | ContractEvent

@step
async def normalize(self, ev: DocumentEvents) -> NormalizedDocumentEvent:
    ...
```

The step decorator already expands unions into the accepted event list, so this keeps routing exact while avoiding duplicate step logic. It does mean the hierarchy is closed/explicit: when a new `DocumentEvent` subtype is added, it also needs to be added to the union.

Would that be an acceptable workaround for your case, or is the pain specifically that the set of subclasses is open/dynamic and hard to enumerate?

vanrajsinh650:
> Thanks, that use case makes sense.
> 
> One possible workaround today is a named union alias for the concrete variants:
> 
> ```python
> DocumentEvents = InvoiceEvent | ReceiptEvent | ContractEvent
> 
> @step
> async def normalize(self, ev: DocumentEvents) -> NormalizedDocumentEvent:
>     ...
> ```
> 
> The step decorator already expands unions into the accepted event list, so this keeps routing exact while avoiding duplicate step logic. It does mean the hierarchy is closed/explicit: when a new `DocumentEvent` subtype is added, it also needs to be added to the union.
> 
> Would that be an acceptable workaround for your case, or is the pain specifically that the set of subclasses is open/dynamic and hard to enumerate?

Thanks, that makes sense. The named union workaround would cover the closed-set case well, and I agree it keeps the current exact routing model non-breaking.

The use case I had in mind is more about the open/extensible hierarchy case, where the base event is part of a reusable workflow or library and downstream users may define additional event subclasses later.

For example:

a reusable workflow exposes DocumentEvent
built-in variants might be InvoiceEvent, ReceiptEvent, and ContractEvent
user code or another package later adds PurchaseOrderEvent(DocumentEvent)
a shared normalization/logging step wants to accept all DocumentEvent variants without needing the original workflow code to be updated every time a new subtype is added

In that case, a union works, but it makes the hierarchy closed and requires every new subtype to be added manually. That is the main pain point I was trying to explore.

That said, I agree the current PR should not change default routing behavior. If you think the union workaround is the preferred pattern for now, I’m okay with closing this PR or converting it into a design discussion. If you think the open-hierarchy case is worth supporting, I can rework the PR into an explicit opt-in mechanism so existing workflows keep exact matching by default.

adrianlyjak:
@vanrajsinh650 that makes sense. I’d be happy with this if we keep exact matching as the default, add a step config to opt into subclass matching, and update the representation/viz path so the graph shows what will actually route at runtime.

vanrajsinh650:
> @vanrajsinh650 that makes sense. I’d be happy with this if we keep exact matching as the default, add a step config to opt into subclass matching, and update the representation/viz path so the graph shows what will actually route at runtime.

Thanks, that makes sense.

I can rework this to keep exact matching as the default and make subclass matching opt-in through step config.

My understanding of the new behavior would be:

default step behavior remains exact-type matching
only steps that explicitly opt in to subclass matching use isinstance / issubclass
validation and build_step_graph follow the same opt-in behavior so the representation matches runtime routing
existing workflows should remain unchanged

I update the PR in that direction.

vanrajsinh650:
Updated the PR to make subclass-aware routing opt-in instead of changing the default behavior. @adrianlyjak 

Summary of changes:

- **API configuration:** Added `accept_event_subclasses: bool = False` to `StepConfig` and the `@step` decorator, and propagated it to `InternalStepConfig`.
- **Default behavior:** Kept normal `@step` routing as exact-type matching.
- **Opt-in subclass routing:** Enabled subclass-aware routing only for steps using `@step(accept_event_subclasses=True)`.
- **Runtime routing and dispatch:** Updated `control_loop.py` and targeted dispatch validation in `workflow.py` to check `accept_event_subclasses` and use `isinstance` only when enabled.
- **Waiter matching:** Scoped waiter matching to the owner step’s opt-in setting.
- **Validation and representation:** Updated `build_step_graph` and `_validate_event_connectivity` so static graph behavior matches runtime routing.
- **Tests:** Added/updated subclass routing tests covering default exact matching, opt-in subclass routing, graph edges, manual targeted validation, and waiter routing.
- **Changeset:** Added a minor changeset for `llama-index-workflows`.

Validation:

- `uv run dev -p workflows -- -k subclass` (10 passed)
- `uv run dev -p workflows` (612 passed)
- `uv run ruff check packages/llama-index-workflows` (passed)
- `uv run ruff format --check packages/llama-index-workflows`  (passed)

vanrajsinh650:
@adrianlyjak hello sir i updated pr can you please review it

vanrajsinh650:
@adrianlyjak  sir can you just take a look at my commits

adrianlyjak:
@vanrajsinh650 have you tested the visualization tools I mentioned earlier? This looks unfinished still. 

Look around here: https://github.com/run-llama/llama-agents/blob/main/packages/llama-index-workflows/src/workflows/representation/build.py#L155

vanrajsinh650:
> @vanrajsinh650 have you tested the visualization tools I mentioned earlier? This looks unfinished still.
> 
> Look around here: https://github.com/run-llama/llama-agents/blob/main/packages/llama-index-workflows/src/workflows/representation/build.py#L155

Good catch. I focused on runtime routing, validation, waiters, and graph connectivity, but I haven't yet updated build.py to reflect accept_event_subclasses.

I update the visualization builder so workflow representations mirror the same opt-in subclass routing behavior and add representation tests to cover it.

Thanku so much for reviewing it.

vanrajsinh650:
> > @vanrajsinh650 have you tested the visualization tools I mentioned earlier? This looks unfinished still.
> > Look around here: https://github.com/run-llama/llama-agents/blob/main/packages/llama-index-workflows/src/workflows/representation/build.py#L155
> 
> Good catch. I focused on runtime routing, validation, waiters, and graph connectivity, but I haven't yet updated build.py to reflect accept_event_subclasses.
> 
> I update the visualization builder so workflow representations mirror the same opt-in subclass routing behavior and add representation tests to cover it.
> 
> Thanku so much for reviewing it.

Thanks @adrianlyjak  for pointing that out. You were right that my earlier implementation only updated runtime routing and validation, but not the workflow representation layer.

I've now updated `build.py` so the visualization follows the same opt-in semantics as runtime execution:

- Default steps continue to use exact-type matching in the generated graph.
- Steps using `@step(accept_event_subclasses=True)` now generate edges from matching subclass event nodes.
- Added representation tests covering:
  - opt-in subclass visualization,
  - default exact-matching behavior,
  - generic annotation safety (`Dict[str, Any]`, `List[str]`, etc.).

Validation:
- `uv run dev -p workflows -- -k subclass`  (11 passed)
- `uv run dev -p workflows -- -k representation`  (34 passed)
- `uv run dev -p workflows`  (615 passed)

The representation graph now matches the runtime routing behavior for opted-in steps.

adrianlyjak:
@vanrajsinh650 thanks for the contribution! I pushed some minor cleanup to remove duplication and fix some minor edge conditions in the visualizations. I'll merge this in shortly, and it will go out with our next release

adrianlyjak (approved):

vanrajsinh650:
Thanks so much for the review, guidance, and the cleanup commits. I learned a lot from the feedback throughout this PR.

I really appreciate you taking the time to explain the concerns, suggest a better direction, and help refine the implementation. Looking forward to contributing more in the future.

Thanks again!

PR #658, opened 2026-06-01
Probe Coveralls coverage path mapping

adrianlyjak:
Disposable probe for Coveralls path mapping. This adds one covered line under the workflows package so the Coveralls comment can show whether changed Python lines resolve after the package base-path change.


adrianlyjak:
Probe complete. Coveralls reported 1/1 changed lines covered, so the package path mapping works for changed Python lines.

PR #674, opened 2026-06-11
Join fan-out streams with list[E] parameters

adrianlyjak:
Workflow steps can now fan out with `list[E]` returns, but on `main` there is no typed way to join the stream back together. Consumers still have to hand-roll buffering, counts, and resume behavior around individual events.

This adds signature-driven collection parameters: a step with a single `list[E]` parameter joins the finite stream opened by an upstream `list[E]` return.

```python
@step
async def fan_out(self, ev: StartEvent) -> list[Task]:
    return [Task(n=i) for i in range(5)]

@step
async def work(self, ev: Task) -> Done:
    return Done(n=ev.n)

@step
async def join(self, events: list[Done]) -> StopEvent:
    return StopEvent(result=[e.n for e in events])
```

Each returned list opens a collection stream. Events emitted inside that stream carry a scope path, and the runtime counts live work until the stream closes. A matching `list[E]` collect parameter then fires once with the collected batch. `Annotated[list[E], Collect(Take(n))]` releases on the nth member, `list[A | B]` collects a flat heterogeneous stream, `return []` still opens and closes an empty stream so joins fire with `[]`, and nested fan-outs create nested stream scopes.

The stream work record is persisted with the existing v2 context shape: queued and in-progress attempts carry scope paths, collect invocations carry their release payload, and open streams/release buffers serialize with the broker state. Retries, catch-error routing, `ctx.wait_for_event`, and snapshot/resume re-deliver the same batch instead of reconstructing it from mutable buffers.

The runtime now fails loudly for collection states that cannot complete: targeted sends into collect steps, stuck open streams at idle, orphaned release buffers, and static graph shapes where a collect step has no compatible returned-list producer or tries to join events from incompatible stream levels.

### DIFF
--- a/packages/llama-index-workflows/src/workflows/context/context_types.py
+++ b/packages/llama-index-workflows/src/workflows/context/context_types.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 from typing import Any
 
@@ -13,8 +15,9 @@ MODEL_T = TypeVar("MODEL_T", bound=BaseModel, default=DictState)  # type: ignore
 # Serialization format version.
 #   v0: legacy nested-JSON-string format (SerializedContextV0).
 #   v1: structured format; per-worker ``in_progress`` was a list of event strings.
-#   v2: ``in_progress`` carries full attempts (retry counts and timestamps), so
-#       a resumed run does not silently restart in-flight work from attempt 0.
+#   v2: per-worker ``in_progress`` carries full attempts (retry counts and
+#       timestamps); collection stream attempts carry ``scope_path`` and collect
+#       invocations carry explicit release payloads.
 CURRENT_SERIALIZED_VERSION = 2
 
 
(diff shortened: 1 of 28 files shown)


packages/llama-index-workflows/src/workflows/representation/validate.py
```diff
@@ -634,6 +639,101 @@ class _WorkflowValidationResult:
...
+    start_event_class: type[StartEvent],
+) -> None:
+    """Reject multi-slot joins whose slots cannot fill at one stream level.
+
+    A multi-slot step fires when one event of each declared type has arrived,
+    and its output continues at the scope of its inputs. That is only
+    well-defined when every slot type is producible at a common stream level
+    (the top level, or one returned-list producer's stream). Mixing levels
+    would strand work items: a slot filled at one level waits forever for a
+    slot that only fills at another.
```
adrianlyjak, line 688: this could use a code illustration/example in the comment to make this more readable

packages/llama-index-workflows/src/workflows/context/context_types.py
```diff
@@ -13,8 +15,8 @@
 # Serialization format version.
 #   v0: legacy nested-JSON-string format (SerializedContextV0).
 #   v1: structured format; per-worker ``in_progress`` was a list of event strings.
-#   v2: ``in_progress`` carries full attempts (retry counts and timestamps), so
-#       a resumed run does not silently restart in-flight work from attempt 0.
+#   v2: collection stream state: attempts carry ``scope_path`` and collect
+#       invocations carry explicit release payloads.
```
adrianlyjak, line 20: looks like we dropped the in_progress comment, which is also part of v2 here

PR #691, opened 2026-06-16
Document child workflows and regroup the workflow guide

adrianlyjak:
Two things in the workflows guide.

## Child workflows page

The guide covers looping, branching, concurrency, and state, but nothing on composing one workflow inside another. The child workflows feature adds that, so there's a new `workflows/child_workflows.md` covering:

- declaring a child as a typed class field (`summarize: Summarize`) and the synthesized constructor that asks for it
- the start/stop boundary: a child needs its own `StartEvent`/`StopEvent` subclasses, the parent triggers it by emitting the child's start, and the child's stop comes back as a routable parent event without ending the run
- nesting children, and namespace isolation of both state and event routing (shared event types stay within their namespace)
- streaming, where child events are hidden by default and `stream_events(include_children=True)` surfaces them tagged via `get_event_origin_namespace`
- which run-level config a nested child honors (`timeout`) versus ignores (`verbose`)

This documents the feature on the child workflows branch; it should land with or after that.

## Sidebar regroup

The page orders had grown ad hoc, with two pairs colliding (`client`/`durable_workflows` at 13, `async_workflows`/`observability` at 15) and topics scattered. Regrouped into four conceptual sections, ordered by `sidebar.order` only (no folders, so no redirects):

- **Introduction** (1): the model plus one runnable example
- **Core workflow features** (2-12): branches and loops, concurrent execution, async, streaming, managing state, custom start and stop events, child workflows, resources, unbound functions, error handling, human in the loop
- **Beyond scripting** (13-16): durable workflows, DBOS, drawing, observability
- **Running as a service** (17-18): server, Python client

Also renames the "Customizing entry and exit points" page to "Custom start and stop events", since that's the actual terminology it teaches. Filename is unchanged.

### DIFF
--- a/docs/src/content/docs/llamaagents/workflows/async_workflows.md
+++ b/docs/src/content/docs/llamaagents/workflows/async_workflows.md
@@ -1,6 +1,6 @@
 ---
 sidebar:
-  order: 15
+  order: 4
 title: Writing async workflows
 ---
 
--- a/docs/src/content/docs/llamaagents/workflows/child_workflows.md
+++ b/docs/src/content/docs/llamaagents/workflows/child_workflows.md
@@ -1,6 +1,6 @@
 ---
 sidebar:
-  order: 4
+  order: 5.5
 title: Child workflows
 ---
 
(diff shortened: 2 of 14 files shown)


adrianlyjak:
Split into two focused PRs: child workflows docs page, and the sidebar regroup.

PR #695, opened 2026-06-17
Add isolated in-process child workflow runs

adrianlyjak:
The workflow runtime had namespace-aware state plumbing, but the broker still treated child workflow steps as static slots. A parent could route into a typed child workflow, but repeated calls through the same child field reused the same worker buffers, timeout clock, stream origin, and child state. This branch gives child workflows an in-process run boundary with a separate runtime invocation id for each call.

A parent declares a child as a typed `Workflow` field, emits the child's custom `StartEvent`, and consumes the child's custom `StopEvent` as an ordinary parent event:

```python
class Child(Workflow):
    @step
    async def run_child(self, ev: ChildStart) -> ChildStop:
        return ChildStop(out=ev.payload.upper())

class Parent(Workflow):
    child: Child

    @step
    async def start(self, ev: StartEvent) -> ChildStart:
        return ChildStart(payload="hello")

    @step
    async def finish(self, ev: ChildStop) -> StopEvent:
        return StopEvent(result=ev.out)
```

## Runtime shape

Step configuration stays static: child steps are keyed by `StepId(("child",), "run_child")`, and root step names still stringify exactly as before. Runtime work gets a separate invocation namespace such as `("child#abc123",)`. The reducer uses the static slot namespace for graph validation, handler lookup, and timeout configuration, while queues, in-progress workers, waiters, collect buffers, task cancellation, stream origins, and `ctx.store` use the concrete invocation namespace.

That means two overlapping calls to the same child field now behave like two child runs. One child `StopEvent` tears down only that invocation and its descendants, one child timeout expires only that invocation, and child state from a completed call is not copied into the next call.

## State and streams

Running context snapshots preserve live child invocation state, including idle child invocations that are waiting for an external response. Completed context continuations carry root state only, so stale child invocation stores do not seed later runs.

Child stream events expose the concrete invocation namespace. `handler.stream_events()` still hides child-origin events by default, and `handler.stream_events(include_children=True)` surfaces them with origins like `child#abc123`. Targeted external sends into a child use that concrete path, for example `child#abc123/answer`; a static `child/answer` target fails instead of accidentally running a child step in the root invocation.

## Runtime support boundary

The in-process runtime owns child workflow execution here. Server and DBOS runtimes reject child workflow attachment early until they can persist and replay invocation-scoped child runs deliberately. The durable state-store changes in this branch keep namespace records separate, but they do not imply durable child workflow support yet.

Smaller replacement for #648.

### DIFF
--- a/.changeset/child-workflows.md
+++ b/.changeset/child-workflows.md
@@ -1,0 +1,12 @@
+---
+"llama-index-workflows": minor
+"llama-agents-agentcore": patch
+"llama-agents-server": patch
+"llama-agents-dbos": patch
+---
+
+Add in-process child workflow composition with typed child declarations, namespaced execution, per-child state, catch-error recovery, timeouts, and opt-in child event streaming.
+
+Annotated Workflow attributes are only auto-attached when they use the typed child workflow contract, so existing manual composition with bare StartEvent/StopEvent workflows remains compatible.
+
+Update server, DBOS, and AgentCore runtime adapter compatibility for namespaced step IDs and state-store access.
--- a/.changeset/per-namespace-state-records.md
+++ b/.changeset/per-namespace-state-records.md
@@ -1,0 +1,5 @@
+---
+"llama-agents-server": patch
+---
+
+Store workflow state as one record per (run_id, namespace) instead of a single bundled record per run. Adds an additive migration that introduces a namespace column defaulting to the root namespace, so existing rows are unchanged.
--- a/packages/llama-agents-agentcore/pyproject.toml
+++ b/packages/llama-agents-agentcore/pyproject.toml
@@ -24,7 +24,8 @@ dependencies = [
   "boto3>=1.42.75",
   "llama-agents-core>=0.5.0",
   "llama-agents-appserver>=0.5.0",
-  "llama-agents-server>=0.2.0",
+  "llama-agents-server>=0.6.1",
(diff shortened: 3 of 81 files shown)


.changeset/child-workflows.md
```diff
@@ -0,0 +1,12 @@
+---
```
adrianlyjak, line 1: This is a novel. Keep changes sets brief. They are not documentation

.changeset/per-namespace-state-records.md
```diff
@@ -0,0 +1,5 @@
+---
+"llama-agents-server": patch
+---
```
adrianlyjak, line 3: Effectively duplicate

packages/llama-agents-agentcore/pyproject.toml
```diff
@@ -24,7 +24,8 @@ dependencies = [
   "boto3>=1.42.75",
   "llama-agents-core>=0.5.0",
   "llama-agents-appserver>=0.5.0",
-  "llama-agents-server>=0.2.0",
+  "llama-agents-server>=0.6.1",
+  "llama-index-workflows>=2.22.0,<3.0.0",
```
adrianlyjak, line 28: These need another bump 2.22 was already released

adrianlyjak: And this shouldn’t add workflows, rather should bump via server?

PR #699, opened 2026-06-18
Run durable workflows in process

adrianlyjak:
Durable workflow persistence already exists behind `AbstractWorkflowStore`, but the packaged path that wires handler rows, event recording, tick persistence, restart, and idle release together has been `WorkflowServer`. Local scripts and workers that do not want HTTP either had to build that stack by hand or fall back to manual checkpointing.

This adds an experimental non-HTTP runtime over the same server persistence machinery. The shape is intentionally small: register workflows, start/stop the runtime, start a run with a stable handler id, and later send events to that id. It does not expose local handler lookup or status/query helpers. A resumed workflow can continue, but there is no `WorkflowHandler` to recover from another process.

```python
from pathlib import Path

from llama_agents.server import SqliteWorkflowStore
from llama_agents.server.runtime import _DurableWorkflowRuntime
from workflows.events import HumanResponseEvent, StartEvent, StopEvent


class Approval(HumanResponseEvent):
    response: str


class DraftWorkflow(Workflow):
    @step
    async def create_draft(self, ctx: Context, ev: StartEvent) -> None:
        await ctx.store.set("draft", f"Draft for {ev.get('topic')}")

    @step
    async def approve(self, ctx: Context, ev: Approval) -> StopEvent:
        draft = await ctx.store.get("draft")
        return StopEvent(result=f"{draft}: {ev.response}")


db_path = Path("workflow-state.db")
handler_id = "draft-42"

runtime = _DurableWorkflowRuntime(
    workflow_store=SqliteWorkflowStore(str(db_path)),
)
runtime.add_workflow("draft", DraftWorkflow())
await runtime.start()
await runtime.run("draft", handler_id=handler_id, topic="quarterly update")

# Simulate process death. The active local handler is gone, but the run has
# persisted enough ticks and state for the next runtime to replay it.
await runtime.stop()

runtime = _DurableWorkflowRuntime(
    workflow_store=SqliteWorkflowStore(str(db_path)),
)
runtime.add_workflow("draft", DraftWorkflow())
await runtime.start()
await runtime.send_event(handler_id, Approval(response="approved"))
await runtime.stop()
```

`run()` waits until the run has persisted enough state to be replayed, so immediate shutdown after `run()` is recoverable. `start()` waits for startup resume in the in-process path, while `WorkflowServer` keeps the existing startup grace window for HTTP request races.

`WorkflowServer` now uses the same private durable runtime object internally instead of carrying a second lifecycle core. Custom server runtimes keep the previous startup ordering, and DBOS can still provide its runtime/store pair so the server-style handler/event/tick state is tracked in the database alongside DBOS execution state.

The docs page change was removed. This should stay experimental until the core workflow APIs settle around durable handler identity.


packages/llama-agents-server/src/llama_agents/server/__init__.py
```diff
@@ -9,12 +9,20 @@
...
+)
 from .server import WorkflowServer
 
 __all__ = [
     "AgentDataStore",
     "AbstractWorkflowStore",
+    "DurableWorkflowHandler",
+    "DurableWorkflowRuntime",
     "HandlerQuery",
+    "ManagedWorkflowRuntime",
```
adrianlyjak: this doesn't seem like part of the public API.

packages/llama-agents-server/src/llama_agents/server/runtime.py
```diff
@@ -0,0 +1,489 @@
...
+        step: str | None = None,
+    ) -> None:
+        """Send an event to a persisted running handler."""
+        self._ensure_started()
+        try:
+            await self._service.send_event(handler_id, event, step=step)
+        except EventSendError as exc:
+            raise RuntimeError(str(exc)) from exc
+
+    async def load_active_handler(self, handler_id: str) -> DurableWorkflowHandler:
```
adrianlyjak: I don't understand what this is for Seems to just be for tests? Isn't there a better way to do this via creating a workflow handle passed a context or something?

.changeset/in-process-runtime.md
```diff
@@ -0,0 +1,6 @@
+---
+"llama-agents-server": patch
```
adrianlyjak: this seems like a minor

packages/llama-agents-server/src/llama_agents/server/runtime.py
```diff
@@ -0,0 +1,489 @@
...
+            return
+        assert unsupported_error is not None
+        raise unsupported_error
+
+    def _ensure_started(self) -> None:
+        if not self._started:
+            raise RuntimeError("ManagedWorkflowRuntime is not started")
+
+
+class DurableWorkflowRuntime:
```
adrianlyjak: wait, why does this exist? Can't it just be the managed one?

packages/llama-agents-server/src/llama_agents/server/runtime.py
```diff
@@ -0,0 +1,489 @@
...
+        self,
+        *,
+        workflow_store: AbstractWorkflowStore | None = None,
+        runtime: Runtime | None = None,
+        resume_existing: bool = True,
+        resume_fresh_handler_grace: timedelta | None = None,
+        wait_for_resume: bool = False,
+        idle_timeout: float | None = None,
+        abort_active_on_stop: bool = True,
+        enable_persistence: bool = True,
```
adrianlyjak: isn't this always the case? Otherwise you shouldn't be using this

PR #704, opened 2026-06-22
fix(workflows): keep implicit waiters distinct across fan-out

### LINKED ISSUE #22070: [Bug]: FunctionAgent parallel tool fan-out + ctx.wait_for_event(HumanResponseEvent) collide on a shared waiter_id, hanging all but one branch
### Bug Description

[`FunctionAgent` in `llama-index-core/llama_index/core/agent/workflow/function_agent.py`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-core/llama_index/core/agent/workflow/function_agent.py#L27-L30) defaults `allow_parallel_tool_calls: bool = True`. When the LLM emits N parallel tool calls, [`BaseWorkflowAgent.parse_agent_output`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-core/llama_index/core/agent/workflow/base_agent.py#L613-L620) fans out N concurrent `@step call_tool` invocations via `ctx.send_event(ToolCall(...))`. If any of those tools wraps a human-in-the-loop wait via `ctx.wait_for_event(HumanResponseEvent, ...)` (the documented HITL pattern in the `llama-index-workflows` runtime) *without* passing a unique `waiter_id`, every parallel invocation derives the same waiter key:

```python
# llama-index-workflows: workflows/context/internal_context.py:193-195
waiter_id = waiter_id or f"waiter_{event_type.__module__}.{event_type.__name__}_{requirements_str}"
```

The runtime then collapses identical-id waiters in place (`workflows/runtime/control_loop/reduce.py:633-634`): `worker_state.collected_waiters[existing] = new_waiter`. Only one waiter survives, and only one `InputRequiredEvent` is ever published to clients (the overwrite path does not re-publish `waiter_event`). When a single `HumanResponseEvent` arrives, exactly one branch resumes; the other N-1 hang until timeout.

The bug fires only when the user's HITL tool calls `ctx.wait_for_event(HumanResponseEvent)` without supplying a distinct `waiter_id`. That's the canonical pattern in the runtime docstring, so the failure mode is reachable on default usage, but a HITL tool that derives `waiter_id` from its own tool-call id is unaffected. Graph-native because it requires framework fan-out + scheduler-managed suspend/resume + shared waiter registry — none of these exist in plain Python.

### Version

llama-index-core==0.14.22

### Steps to Reproduce

The PoC mirrors `FunctionAgent`'s fan-out shape inside a `workflows.Workflow` (avoiding the need for a real LLM) and uses the real `ctx.wait_for_event(HumanResponseEvent)` primitive — the documented HITL example in the runtime docstring, which passes no `waiter_id`.

```python
import asyncio
from llama_index.core.workflow import (
    Context, Event, HumanResponseEvent, InputRequiredEvent,
    StartEvent, StopEvent, Workflow, step,
)

class ToolCallEv(Event):  tool_id: str
class ToolDoneEv(Event):  tool_id: str; resumed_with: str
class FanInEv(Event):     pass

N = 3

class ParallelHITLWorkflow(Workflow):
    @step
    async def dispatch(self, ctx: Context, ev: StartEvent) -> ToolCallEv | FanInEv:
        for i in range(N):
            ctx.send_event(ToolCallEv(tool_id=f"tool_{i}"))     # parse_agent_output shape
        await ctx.store.set("expected", N)
        return FanInEv()

    @step(num_workers=N)
    async def call_tool(self, ctx: Context, ev: ToolCallEv) -> ToolDoneEv:
        # Canonical HITL pattern — no waiter_id passed.
        resp = await ctx.wait_for_event(
            HumanResponseEvent,
            waiter_event=InputRequiredEvent(prefix=f"approve {ev.tool_id}? "),
            timeout=4.0,
        )
        return ToolDoneEv(tool_id=ev.tool_id, resumed_with=resp.response)

    @step
    async def aggregate(self, ctx: Context, ev: ToolDoneEv | FanInEv) -> StopEvent | None:
        expected = await ctx.store.get("expected", default=None)
        if expected is None: return None
        results = ctx.collect_events(ev, expected=[ToolDoneEv] * expected)
        if results is None: return None
        return StopEvent(result=[(r.tool_id, r.resumed_with) for r in results])

async def main():
    wf = ParallelHITLWorkflow(timeout=6.0)
    handler = wf.run()
    seen = 0
    async for ev in handler.stream_events():
        if isinstance(ev, InputRequiredEvent):
            seen += 1
            if seen == 1:
                handler.ctx.send_event(HumanResponseEvent(response="APPROVE"))
            break
    try:
        await asyncio.wait_for(handler, timeout=5.0); completed = True
    except Exception: completed = False
    print(f"InputRequiredEvents seen: {seen} of {N}")
    print(f"all branches completed   : {completed}")

asyncio.run(main())
# CONFIRMED BUG when seen < N and completed is False
```

### Relevant Logs/Tracebacks

```shell
``
[1] FunctionAgent.allow_parallel_tool_calls default == True : True
    BaseWorkflowAgent.parse_agent_output fans out via ctx.send_event in a for loop : True

[3] OBSERVED
    seen_input_required_events: 1
    expected_input_required_events: 3
    all_branches_completed: False
    result: TimeoutError:

[sig 1] waiter_event collapse  (IRQs < N): True  (1 of 3)
[sig 2] branches stuck after 1 HRE       : True
```

fengjikui:
## Summary
- include the current trigger event in generated implicit waiter IDs
- keep explicit `waiter_id` behavior unchanged
- add a regression test for parallel branches that all wait for `HumanResponseEvent` with implicit waiter IDs

## Root cause
`Context.wait_for_event()` generated the default waiter ID from only the awaited event type and requirements. When multiple workers handled different fan-out events and waited for the same event type without an explicit `waiter_id`, each branch produced the same default ID, so later waiters overwrote earlier ones in the control loop. Only one `InputRequiredEvent` was emitted and one branch resumed.

The fix adds a stable fingerprint of the step invocation trigger event to implicit waiter IDs, so different fan-out branches get distinct generated IDs while a suspended invocation can recreate the same ID when resumed. Explicit waiter IDs still behave exactly as before.

Fixes run-llama/llama_index#22070.

## Validation
- `uv run ruff format --check packages/llama-index-workflows/src/workflows/context/internal_context.py packages/llama-index-workflows/src/workflows/runtime/types/results.py packages/llama-index-workflows/src/workflows/runtime/types/step_function.py packages/llama-index-workflows/tests/context/test_context.py`
- `uv run ruff check packages/llama-index-workflows/src/workflows/context/internal_context.py packages/llama-index-workflows/src/workflows/runtime/types/results.py packages/llama-index-workflows/src/workflows/runtime/types/step_function.py packages/llama-index-workflows/tests/context/test_context.py`
- `uv run --project packages/llama-index-workflows pytest packages/llama-index-workflows/tests/context/test_context.py::test_parallel_implicit_waiters_are_not_collapsed -q`
- `uv run --project packages/llama-index-workflows pytest packages/llama-index-workflows/tests/context/test_context.py::test_wait_for_multiple_events_in_workflow packages/llama-index-workflows/tests/context/test_context.py::test_to_dict_after_stream_events_break_resumes -q`
- `uv run --project packages/llama-index-workflows pytest packages/llama-index-workflows/tests/runtime/test_control_loop.py -k 'wait_for_event or InputRequiredEvent' -q`
- `uv run --project packages/llama-index-workflows pytest packages/llama-index-workflows/tests/context/test_context.py -q`
- `git diff --check`

### DIFF
--- a/.changeset/implicit-waiter-fanout.md
+++ b/.changeset/implicit-waiter-fanout.md
@@ -1,0 +1,5 @@
+---
+"llama-index-workflows": patch
+---
+
+Keep implicit waiters distinct across parallel workflow fan-out branches.
--- a/packages/llama-index-workflows/src/workflows/context/context_types.py
+++ b/packages/llama-index-workflows/src/workflows/context/context_types.py
@@ -100,6 +100,9 @@ class SerializedEventAttempt(BaseModel):
     # Explicit collect invocation payload, serialized only for queued/in-progress
     # list[E] collect executions.
     collection_release_payload: SerializedCollectionReleasePayload | None = None
+    # Stable identity of this queued or in-progress work item. Additive:
+    # older payloads validate with None and get fresh ids on requeue.
+    work_item_id: str | None = None
 
 
 class SerializedCollectionReleasePayload(BaseModel):
@@ -131,6 +134,8 @@ class SerializedWaiter(BaseModel):
     scope_path: list[str] = Field(default_factory=list)
     # For a suspended collect invocation, the release batch to re-invoke with.
     collection_release_payload: SerializedCollectionReleasePayload | None = None
+    # Stable identity for the suspended work item, if available.
+    work_item_id: str | None = None
 
     @model_validator(mode="before")
     @classmethod
@@ -203,6 +208,9 @@ class SerializedContext(BaseModel):
     # Monotonic stream-id counter. Persisted so a resumed run keeps minting
     # unique, deterministic stream ids.
     stream_seq: int = Field(default=0)
+    # Monotonic work-item counter. Persisted so implicit waiter ids remain
+    # stable and unique across serialize/resume boundaries.
+    work_item_seq: int = Field(default=0)
     streams: dict[str, SerializedCollectionStreamInstance] = Field(default_factory=dict)
     collection_release_states: dict[str, SerializedCollectionReleaseState] = Field(
         default_factory=dict
--- a/packages/llama-index-workflows/src/workflows/context/internal_context.py
+++ b/packages/llama-index-workflows/src/workflows/context/internal_context.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import json
 import logging
 import time
 from collections import Counter, defaultdict
@@ -107,6 +108,37 @@ class InternalContext(Generic[MODEL_T]):
                 f"{fn} may only be called from within a step function"
             )
 
+    @classmethod
+    def _stable_requirements_str(cls, requirements: dict[str, Any]) -> str:
+        """Return a deterministic representation for waiter requirements."""
+
+        def normalize(value: Any) -> Any:
+            if isinstance(value, dict):
+                return {
+                    str(k): normalize(v)
+                    for k, v in sorted(value.items(), key=lambda item: repr(item[0]))
+                }
+            if isinstance(value, (list, tuple)):
+                return [normalize(v) for v in value]
+            if isinstance(value, set):
+                return sorted((normalize(v) for v in value), key=repr)
+            if hasattr(value, "model_dump"):
+                try:
+                    return normalize(value.model_dump())
+                except Exception:
+                    pass
+            return value
+
+        try:
+            return json.dumps(
+                normalize(requirements),
+                sort_keys=True,
+                separators=(",", ":"),
+                default=repr,
+            )
+        except Exception:
+            return repr(requirements)
+
     @property
     def store(self) -> StateStore[MODEL_T]:
         """Access workflow state store."""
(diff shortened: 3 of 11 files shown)


adrianlyjak:
@fengjikui good call out. There's a lot of complexity here unfortunately, and I'm not sure this is complete and there's a fair amount to think about here

A few concerns I have with this:
- waiters without explicit waiter ids (defaulting before this is rolled out, and then serialized), and then resumed, will no longer match. In fact this will fire a new input required event
- identical events will still have the same failure scenario. Fan outs could easily do this to do trigger work, either pulling work units from external state, or just doing same shaped work
- light chance of non-determinism across serialization (set/dict ordering). Especially since this crosses serialization boundaries.

fengjikui:
Thanks for spelling these out. I pushed `15bbbbc` to address the waiter identity cases directly:

- implicit waiters now carry a stable `work_item_id` through queued/in-progress/waiter serialization, so parallel fan-out work items do not collapse even when the triggering events have the same shape
- serialized payloads that predate `work_item_id` still fall back to the legacy implicit waiter id, so an existing suspended waiter can resume instead of firing a fresh input-required event
- implicit waiter requirement strings now use a deterministic normalized JSON representation, including sorted dict/set handling, before falling back to `repr`

I added regression coverage for the legacy resume case, distinct parallel branches, and identical parallel events. Local validation just now:

```
uv run dev -p workflows -- tests/context/test_context.py -q -k "waiter or parallel_implicit or legacy_implicit"
# 3 passed
uv run ruff check packages/llama-index-workflows/src/workflows/context/context_types.py packages/llama-index-workflows/src/workflows/context/internal_context.py packages/llama-index-workflows/src/workflows/runtime/control_loop/reduce.py packages/llama-index-workflows/src/workflows/runtime/types/internal_state.py packages/llama-index-workflows/src/workflows/runtime/types/results.py packages/llama-index-workflows/src/workflows/runtime/types/step_function.py packages/llama-index-workflows/src/workflows/runtime/types/ticks.py packages/llama-index-workflows/tests/context/test_context.py
# All checks passed
```

adrianlyjak:
Pushed a commit on top: **collect (`list[E]`) invocations now get a stable work item id.**

The counter-based `work_item_id` is minted in `_process_add_event_tick`, but collect invocations are fired directly from the stream-close path (`_fire_collection_release`) rather than through a routed tick, so they never got one — their implicit waiter fell back to the unsuffixed base id. Two collect invocations of the same join step (one per fan-out source) then collapse onto a single waiter, and only one resumes:

```python
@step
async def fan_a(self, ev: StartEvent) -> list[Done]: ...
@step
async def fan_b(self, ev: StartEvent) -> list[Done]: ...
@step
async def join(self, ctx: Context, events: list[Done]) -> Batch:
    await ctx.wait_for_event(HumanResponseEvent, waiter_event=InputRequiredEvent())
    ...
```

The fix derives a collect invocation's id from its `stream+binding` key (`CollectionReleasePayload.work_item_id()`), which is already unique and carried on the payload across serialize/resume. It's set at the release site and in the per-tick mint, so re-delivery recreates the same id. Added a regression test in `test_collection_stream_regressions.py` that hangs without the fix.

adrianlyjak:
@fengjikui thanks for the PR!

PR #712, opened 2026-06-30
Handle AgentWorkflow memory in in-memory state edits

### LINKED ISSUE #709: AgentWorkflow handoff regression after 2.20.0 in streamed multi-agent chat
Hi team, we may be hitting a regression in `llama-index-workflows` after `2.20.0` when using `AgentWorkflow` handoff in a streamed multi-agent chat setup.

This does **not** look like a terminal / CLI / logging issue on our side.
The failure reproduces specifically through **agent handoff flow**.

## Current status

We currently pin:

- `llama-index-workflows==2.20.0`

because newer versions (`>=2.21.x`, and also reproduced on `2.22.1`) appear to break our handoff behavior.

## Our usage pattern

We use `AgentWorkflow` with a router agent that can hand off to a member/specialist agent.

Simplified flow:

1. Router agent receives the user message
2. Router calls `handoff`
3. Workflow switches to member agent
4. Member agent should stream the actual final response

This worked for us on `2.20.0`.

## What breaks on newer versions

After upgrading beyond `2.20.0`, the handoff flow appears to silently fail in practice:

- router receives the message
- handoff is triggered
- but customer-facing stream never yields the downstream member response correctly
- UI ends up with fallback / empty response behavior instead of the specialist answer

## Expected behavior

After router emits a handoff to a member agent:

- we should see the member agent become active
- member agent stream/delta events should continue
- final response should come from the member agent

## Actual behavior

On newer versions:

- handoff-related events may occur internally
- but downstream member streaming/result does not surface as expected in the application flow
- end-user effectively gets no proper specialist response

## Minimal event sequence we rely on

Conceptually, this sequence should hold:

1. `AgentInput(current_agent=router)`
2. `ToolCall(tool_name=handoff)`
3. `ToolCallResult(tool_name=handoff)`
4. `AgentInput(current_agent=member)`
5. `AgentStream(current_agent=member, delta=...)`

On `2.20.0`, this behavior is stable for us.
On newer versions, the handoff path seems to no longer behave compatibly.

adrianlyjak:
AgentWorkflow stores the public `Memory` object under the `memory` state key before it initializes the rest of the context. On main, the in-memory state store deep-copies the whole `DictState` for every edit, so `workflow.run(..., memory=Memory.from_defaults(...))` can recurse into the live chat store and fail while copying objects that are not meant to be copied.

This keeps edit isolation for ordinary `DictState` entries, but treats known unserializable keys the same way serialization already does. The in-memory edit copy preserves those live objects by reference and deep-copies the rest. The new integration coverage runs an `AgentWorkflow` with `llama_index.core.memory.Memory`; the state-store coverage keeps nested mutable isolation in place while proving the `memory` key can hold a non-deepcopyable object.

Fixes https://github.com/run-llama/llama-agents/issues/709

### DIFF
--- a/packages/llama-agents-integration-tests/tests/test_context_store.py
+++ b/packages/llama-agents-integration-tests/tests/test_context_store.py
@@ -10,7 +10,23 @@ from llama_agents_integration_tests.helpers import (
     make_text_response,
     make_tool_call_response,
 )
+from llama_index.core.base.llms.types import ChatMessage, MessageRole
+from llama_index.core.memory import Memory
 from workflows import Context
+
+
+async def test_agent_workflow_accepts_memory_object(
+    create_workflow: WorkflowFactory,
+) -> None:
+    """Test that workflow.run accepts the public Memory implementation."""
+    workflow = create_workflow(responses=[make_text_response("Done")])
+    memory = Memory.from_defaults(
+        chat_history=[ChatMessage(role=MessageRole.USER, content="earlier turn")]
+    )
+
+    result = await workflow.run(user_msg="hi", memory=memory)
+
+    assert result.response.content == "Done"
 
 
 async def test_initial_state_accessible_in_tool(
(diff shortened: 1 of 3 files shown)


adrianlyjak:
Closing this. The requested next artifact was a regression test/repro to think through the issue, not a fix PR.

PR #719, opened 2026-07-04
AgentDataClient retries transient read failures

kiedanski:
AgentDataClient used the default httpx timeout on its shared client, so a slow Agent Data search could fail after 5 seconds as a `ReadTimeout` with little useful message. Cloud persistence reads through this path, and search is the only idempotent Agent Data operation in this client.

This branch gives the shared client a 30 second read timeout with a 5 second connect timeout, then retries transient search failures with bounded exponential backoff. The retry path covers transport failures and 500/502/503/504 responses. Create, update, and delete paths still make one attempt because Agent Data writes do not have idempotency keys; replaying an ambiguous write timeout could duplicate or overwrite data.

The retry code keeps timeout and attempt config type-stable, validates attempt counts, and has server coverage for retryable reads, final failure, 4xx handling, and all write methods as non-retried. It also adds a patch changeset for `llama-agents-server`.

### DIFF
--- a/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_client.py
+++ b/packages/llama-agents-server/src/llama_agents/server/_store/agent_data_client.py
@@ -23,6 +23,14 @@ _RETRY_BACKOFF_BASE = 0.5
 _RETRYABLE_STATUS = frozenset({500, 502, 503, 504})
 
 
+def _normalize_timeout(timeout: httpx.Timeout | float | None) -> httpx.Timeout:
+    if timeout is None:
+        return _DEFAULT_TIMEOUT
+    if isinstance(timeout, httpx.Timeout):
+        return timeout
+    return httpx.Timeout(timeout)
+
+
 class AgentDataClient:
     """HTTP client for the LlamaCloud Agent Data API.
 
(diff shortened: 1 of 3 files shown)


packages/llama-agents-server/tests/server/test_agent_data_client.py
```diff
@@ -0,0 +1,101 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 LlamaIndex Inc.
+"""Retry/timeout behavior for AgentDataClient."""
```
adrianlyjak: This should go with the existing tests

operator/internal/controller/resources.go
```diff
@@ -1224,6 +1224,9 @@ http {
   sendfile        on;
   keepalive_timeout  65;
   access_log    off;
+  # Match the LlamaCloud ingress default so large uploads aren't rejected
```
adrianlyjak: Remove this comment, not a stable target

packages/llama-agents-server/src/llama_agents/server/_store/agent_data_client.py
```diff
@@ -67,6 +80,40 @@ async def close(self) -> None:
             await self._shared_client.aclose()
             self._shared_client = None
 
+    async def _request(self, method: str, url: str, **kwargs: Any) -> httpx.Response:
+        """Issue a request, retrying transient failures with exponential backoff.
```
adrianlyjak: This is a bad idea for non idempotent requests. Will lead to duplicate workflows, duplicate events. Idempotency needs to be figured out on the backend first

operator/internal/controller/resources.go
```diff
@@ -1224,6 +1224,9 @@ http {
...
+  # at the file-server sidecar before reaching the python app.
+  client_max_body_size 300m;
```
adrianlyjak: Llama agents doesn’t to any file management, so I’m unclear why this is being added

adrianlyjak (approved):

PR #734, opened 2026-08-11
fix(retry): correct attempt-index off-by-one between wait strategies

### LINKED ISSUE #733: Wait strategies silently skip their first configured delay due to attempt-index off-by-one
`RetryPolicy.next()` is always called by the runtime with `attempts`
counting completed failures starting at **1** (see
`runtime/control_loop/reduce.py`: `failures = this_execution.attempts + 1`).

Every attempt-indexed wait strategy (`wait_chain`, `wait_exponential`,
`wait_incrementing`, `wait_exponential_jitter`, `wait_random_exponential`)
is documented and unit-tested as **0-indexed** — index 0 is meant to be
the delay before the first retry.

`_ComposableRetryPolicy.next()` forwards the 1-indexed `attempts` straight
to the wait strategy, so every strategy's first configured delay is
silently skipped, and every later delay fires one step early.

Repro (`wait_chain(wait_fixed(1), wait_fixed(2), wait_fixed(5))` through a
real Workflow retry loop):

    Expected gaps: 1.0s, 2.0s, 5.0s
    Actual gaps:   2.0s, 5.0s, 5.0s

Same shift reproduces with `wait_exponential(multiplier=1, exp_base=2)`:
expected 1.0/2.0/4.0s, actual 2.0/4.0/8.0s.

I have a fix + regression test ready and will open a PR referencing this issue.

Sohel2309:
## Summary

Fixes an off-by-one error in retry wait-strategy indexing.

## Problem

`_ComposableRetryPolicy.next()` forwarded the runtime's 1-indexed `attempts` value (which always starts at 1 on the first failure, see `runtime/control_loop/reduce.py: failures = this_execution.attempts + 1`) straight to the configured wait strategy.

Every attempt-indexed wait strategy (`wait_chain`, `wait_exponential`, `wait_incrementing`, `wait_exponential_jitter`, `wait_random_exponential`) is documented and unit-tested as 0-indexed, where index 0 is the delay before the first retry.

The mismatch meant every wait strategy silently skipped its first configured delay and used the second delay for the first retry, third for the second, and so on. For `wait_chain` specifically, this meant the first strategy in the chain was never used at all.

## Solution

Subtract 1 from `attempts` before passing it to the wait strategy, so it receives the 0-indexed value its contract expects.

Stop conditions are unaffected and continue to receive the original 1-indexed `attempts`, which is the correct convention for `stop_after_attempt`.

The existing unit tests for `.next()` were also updated because they called it directly with attempts starting at 0, an assumption that did not match how the runtime actually invokes it. A regression test was added that reproduces the real retry-loop calling convention end-to-end and fails with the previous behavior.

## Tests

- Full test suite: 857 passed
- Ruff formatting/lint checks passed
- Type checking passed

Fixes #733

### DIFF
--- a/.changeset/retry-wait-indexing.md
+++ b/.changeset/retry-wait-indexing.md
@@ -1,0 +1,5 @@
+---
+"llama-index-workflows": patch
+---
+
+Fix an off-by-one that made retry wait strategies skip their first configured delay, so every retry delay is now one step shorter than before.
--- a/docs/src/content/docs/llamaagents/workflows/retry_steps.md
+++ b/docs/src/content/docs/llamaagents/workflows/retry_steps.md
@@ -162,6 +162,7 @@ def next(
 ```
 
 Return the number of seconds to wait before retrying, or `None` to stop.
+`attempts` counts failures so far, starting at 1 for the first failure.
 The optional `seed` is used by durable runtimes to make jitter deterministic during replay.
 
 For example, this policy only retries on Fridays:
--- a/packages/llama-index-workflows/src/workflows/retry_policy.py
+++ b/packages/llama-index-workflows/src/workflows/retry_policy.py
@@ -90,7 +90,8 @@ class RetryPolicy(Protocol):
 
         Args:
             elapsed_time: Seconds since the first failure.
-            attempts: Number of attempts made so far.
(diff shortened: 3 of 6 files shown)


adrianlyjak:
@Sohel2309 thanks for the PR

PR #737, opened 2026-08-18
Add handler completion polling to WorkflowClient

Whxuan0701:
## Summary

- add `WorkflowClient.wait_for_handler()` for polling asynchronous workflow runs until they reach a terminal state
- support bounded waits with monotonic timeout accounting and configurable polling intervals
- preserve failed and cancelled terminal results while propagating HTTP errors unchanged
- cover completion transitions, terminal states, timeout boundaries, invalid options, and missing handlers
- add a patch changeset for `llama-agents-client`

## Validation

- `uv run pytest packages/llama-agents-client/tests -n0` (62 passed)
- `uv run ruff format --check packages/llama-agents-client/src/llama_agents/client packages/llama-agents-client/tests`
- `uv run ruff check packages/llama-agents-client/src/llama_agents/client packages/llama-agents-client/tests`
- `uv run ty check packages/llama-agents-client/src/llama_agents/client packages/llama-agents-client/tests`
- `git diff --check origin/main...HEAD`

## Known gaps

The full-repository pre-commit run could not complete locally because the workspace environment does not include optional AgentCore and llamactl dependencies. The checks scoped to the affected client package pass.

### DIFF
--- a/packages/llama-agents-client/src/llama_agents/client/client.py
+++ b/packages/llama-agents-client/src/llama_agents/client/client.py
@@ -26,6 +26,7 @@ from .protocol import (
     SendEventResponse,
     Status,
     WorkflowsListResponse,
+    is_status_completed,
 )
 from .protocol.serializable_events import (
     EventEnvelope,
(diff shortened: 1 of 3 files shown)


packages/llama-agents-client/src/llama_agents/client/client.py
```diff
@@ -509,6 +510,50 @@ async def get_handler(self, handler_id: str) -> HandlerData:
 
             return HandlerData.model_validate(response.json())
 
+    async def wait_for_handler(
```
adrianlyjak, line 527: I'd like this to have a back-off by default, otherwise it can be quite a lot of requests for a long workflow. And if we make that change later, it'll be a breaking change, so I think we need that in from the start.

Something very similar to what we have elsewhere: https://github.com/run-llama/llama-parse-py/blob/main/src/llama_cloud/_polling.py#L47

Whxuan0701: Implemented in 4a7f6b1. wait_for_handler now defaults to linear backoff from 0.5 seconds up to a 5-second cap while preserving the immediate first status request and timeout-bounded sleeps. Callers can select constant or exponential backoff and configure the maximum interval. Regression coverage verifies all three strategies, the default cap, and invalid parameter handling. The full client suite passes with 69 tests, and the repository-wide pre-commit checks pass, including ty and basedpyright.

PR #738, opened 2026-08-18
Reject concurrent reuse of active handler IDs

Whxuan0701:
## Summary

- reject HTTP workflow starts that reuse a currently active handler ID
- return `409 Conflict` with a stable, actionable error from both blocking and non-blocking run endpoints
- keep terminal handler ID reuse intact for existing state-handoff workflows
- add endpoint regression coverage and a patch changeset for `llama-agents-server`

## Validation

- `uv run pytest packages/llama-agents-server/tests -n0` (447 passed, 32 deselected)
- `uv run pytest packages/llama-agents-appserver/tests -n0` (107 passed)
- `uv run ruff format --check packages/llama-agents-server/src/llama_agents/server packages/llama-agents-server/tests/server`
- `uv run ruff check packages/llama-agents-server/src/llama_agents/server packages/llama-agents-server/tests/server`
- `uv run ty check packages/llama-agents-server/src/llama_agents/server/_service.py packages/llama-agents-server/src/llama_agents/server/_api.py packages/llama-agents-server/tests/server/test_server_endpoints.py`
- `git diff --check origin/main...HEAD`

## Known gaps

The store lookup and handler creation are not a cross-process compare-and-set operation. This change closes the existing HTTP service path mismatch and does not claim distributed atomicity.

The full-repository pre-commit run could not complete locally because the workspace environment does not include optional AgentCore, Control Plane, and llamactl dependencies. The checks scoped to the affected server and appserver packages pass.

### DIFF
--- a/.changeset/reject-active-handler-reuse.md
+++ b/.changeset/reject-active-handler-reuse.md
@@ -1,0 +1,5 @@
+---
+"llama-agents-server": patch
+---
+
+Reject concurrent HTTP runs that reuse an active workflow handler ID
--- a/packages/llama-agents-server/src/llama_agents/server/_api.py
+++ b/packages/llama-agents-server/src/llama_agents/server/_api.py
@@ -40,6 +40,7 @@ from workflows.utils import _nanoid as nanoid
 
 from ._service import (
     EventSendError,
+    HandlerAlreadyRunningError,
     HandlerCompletedError,
     HandlerNotFoundError,
     _WorkflowService,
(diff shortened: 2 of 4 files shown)


adrianlyjak:
Thanks, this is great. I've been meaning to add this. Another related functionality to this is to be able to specify a run ID for idempotency. In that case, I think we can just return some sort of 200 response if the run ID already exists and skip the 409. But this is a great first step .

PR #749, opened 2026-08-22
fix: use immutable defaults for read-only sequences

harshadkhetpal:
Three read-only parameters use list literals as defaults, so Python shares mutable objects across calls. Their annotations now accept `Sequence` values and their defaults are immutable tuples:

```python
args: Sequence[str] = ()
registry: Sequence[type[Event]] = ()
persistence_backoff: Sequence[float] = (0.5, 3.0)
```

`run_uv` unpacks `args` into the command list, so callers can pass lists or tuples. `WorkflowServer` still copies the retry delays into runtime-owned lists, and an empty sequence still disables retries.

### DIFF
--- a/.changeset/quiet-defaults-fix.md
+++ b/.changeset/quiet-defaults-fix.md
@@ -1,0 +1,7 @@
+---
+"llama-agents-appserver": patch
+"llama-agents-client": patch
+"llama-agents-server": patch
+---
+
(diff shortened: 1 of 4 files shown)


adrianlyjak:
@harshadkhetpal thanks for the PR. these are all non mutated configurations, instead switched to immutable default args to keep it safe rather than having to introduce a None

